PR #6, opened 2018-12-17
Fix self update on windows

### LINKED ISSUE #5: Auto-update fails
Running on windows-amd64, the auto-update fails for me.

azbrowse notifies me that an update is available, but doesn't apply the update (the UI loads, but the old version).

stuartleeks:
On Windows the input from the update prompt included CRLF not just LF, updated to handle this (fixes #5)

Also converted `log.Println` messages in self update error path to `log.Panicln` to make the output easier to see. This is a separate commit that can be reverted if you prefer the old behaviour.


selfupdate.go
```diff
@@ -3,48 +3,44 @@ package main
...
-	"log"
-	"os"
 )
 
 func confirmAndSelfUpdate() {
 	latest, found, err := selfupdate.DetectLatest("lawrencegripper/azbrowse")
 	if err != nil {
-		log.Println("Error occurred while detecting version:", err)
-		return
+		log.Panicln("Error occurred while detecting version:", err)
```
lawrencegripper: I think this would mean the util wouldn't be usable if github was down which feels bad. One option is that we could use a 'ReadString' after the original 'Println' for the user to acknowledge the failure and ensure the log lines aren't lost. 

What do you think?

selfupdate.go
```diff
@@ -3,48 +3,44 @@ package main
...
 	v, err := semver.Parse(version.BuildDataVersion)
 	if err != nil {
-		log.Println(err.Error())
-		return
+		log.Panicln(err.Error())
 	}
 	if !found || latest.Version.LTE(v) {
-		log.Println("Current version is the latest")
-		return
+		log.Panicln("Current version is the latest")
```
lawrencegripper: I think this cause a crash when you have the latest version installed and a check is run as, if the check finds you have latest this `panicln` would run

selfupdate.go
stuartleeks: Ah, good point!

selfupdate.go
stuartleeks: Oops - trigger-happy. I'll revert this one!

PR #16, opened 2018-12-19
WIP Attempt a building a crawler to populate search 

### LINKED ISSUE #11: Feature: Provide search over resources
I'd like to start caching resource names in the Boltdb cache then use this [fuzzy search](https://github.com/sajari/fuzzy) to allow quick searching of resources. 

Currently in two minds as to whether the app should *crawl* all resources to build the cache or just only search over those that have been viewed in the past.

lawrencegripper:
Closes #11

### DIFF
--- a/actions.go
+++ b/actions.go
@@ -25,8 +25,12 @@ func LoadActionsView(list *ListWidget) error {
 	for _, resOps := range opsRequest.ResourceTypes {
 		if resOps.Name == strings.Split(list.CurrentItem().armType, "/")[1] {
 			for _, op := range resOps.Operations {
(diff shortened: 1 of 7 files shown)


lawrencegripper:
It's slow. Options:

- Use lots of go routines
- Look at using the portal api for search
- See if some kind of use of $expand can reduce the number of calls

lawrencegripper:
Thanks to work by @stuartleeks we've got a much more efficient way to crawl (1 call per 1000 resources). This is now ready to be integrated for testing on the command line. Testers can use `azbrowse --search queryTermHere` and see how long the process takes to build the index and how relevant the results are.

Issue #22, opened 2018-12-27
Technical Debt and ongoing dev


lawrencegripper:
As the project moves past an MVP I think it would be good to address the test coverage in the codebase. 

Todos:

 - [x] Add logging functions to write to file
 - [ ] Unit test functions where appropriate
 - [ ] Look at integration testing tools which can replay commands to the running tool to exercise common paths. For example: load -> open sub -> open RG -> list keys for storage account
 - [ ] Document some of the 'hacks' in the codebase better. For example, using the `Microsoft.Authorisation` REST API to rebuild actions URLs for resources

lawrencegripper:
Logging completed in #23

lawrencegripper:
Now that the `treeview` has been refactored I'd like to build out some integration tests around the handlers. Plan to spec this out a bit more over the next week or so.

PR #23, opened 2018-12-27
Using OpenTracing and AppDash to enable logging

lawrencegripper:
Related #22 

Looking at using OpenTracing to provide a tracing/logging interface into the app


lawrencegripper:
Here is a quick example, the server runs at http://localhost:8700

screenshot from 2018-12-27 21-38-48

PR #25, opened 2019-01-04
UI Refresh: Titles, First Run Experience, Help Page, Seperators and Refresh

### LINKED ISSUE #9: Resize ASCII art to priorities keyboard bindings
Just a nit to reduce or restyle the ASCII art to improve the legibility of the keyboard bindings. A screenshot below shows my screen. The description of the keys is missing as it's been rendered off the side of the panel. It'd be a better UX to move this around a little bit to prioritize the ASCII art (Sorry I know it's really nice). 
azbrowse2

### LINKED ISSUE #17: Feature: Refresh
Currently if you want to refresh a list of resource in a resource group or your subscriptions you have to back up and re select the option. It'd be great to have a simple shortcut (F5?) to refresh the list in place.

### LINKED ISSUE #20: Enhancement: Add seperator between resources
I think from a UX POV it'd be better to add a small separator between the resources so it is clearer which resource is currently selected. See below.

sep

lawrencegripper:
Fixes #17 Fixes #20 Fixes #9 

@jjcollinge - As all of these are your requests you happy to have a play and see what you think?


jjcollinge:
Thanks, will do later!

lawrencegripper:
Ps. Original message made me sound a bit ungrateful - all good requests and I think it looks way better with the changes

jjcollinge:
I think this looks and feels great. It'd be nice if there was some icon/indicator when the CLI is doing a remote call i.e. some actions can take a little time, people have variable internet speeds, etc. people might repeat key presses if they don't have some sort of immediate UI response. I think a simple icon would be sufficient to let people know that something is happening.

lawrencegripper:
Thanks for taking a look, I've added an egg timer to indicate in-flight requests

PR #28, opened 2019-01-10
Fix #27 Help toggling also triggering back

### LINKED ISSUE #27: Bug: Showing help resets navigation
### Steps to repro:

1. Navigate to a subscription and resource group (or resource)
2. Press Ctrl+H to show the help page
3. Press Ctrl+H again to dismiss the help page

### Expected behaviour:
The previously selected resource is still displayed

### Actual behaviour:
The starting subscription list is displayed

lawrencegripper:
This was a super weird one. 

1. I had two "ctrl+H" registered (copy and paste error). 
2. "keyBackspace" and "keyCtrlH" did the same thing (I've removed keybackspace and now just use keybackspace2)

While I was messing around right and left arrows now go back and forward in the navigation stack too

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ before_deploy:
 - export BODY=$(git log -1  --pretty='%s')
 
 deploy:
+  skip_cleanup: true
   provider: releases
   api_key:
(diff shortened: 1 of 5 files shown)


lawrencegripper:
@stuartleeks you ok to give this a text on your windows box and make sure I haven't messed anything up on windows?

stuartleeks:
@lawrencegripper Ctrl+H is working fine (i.e. leaving you in the current resource as you toggle help), but Backspace is no longer navigating back up the resource hierarchy (which I found useful)

lawrencegripper:
@stuartleeks good spot it turns out that on older terminals Backspace is `^H` which means it clashes. I've moved the help to `CTRL+I` to avoid the issue while also supporting the widest range of terminals... not ideal but I think it should do the trick. 

I also tried to make some of the features more discoverable by adding keyboard shortcuts to the title bars when they're available.

PR #35, opened 2019-02-06
Updates to handler refactoring

stuartleeks:
Fix uncaptured loop variable that caused the last defined handler to always be used
Extend handler timeout to 10 mins when `debug` flag is specified


lawrencegripper:
Awesome work! Thanks!

PR #40, opened 2019-02-06
Resource status using Azure resource Graph & Folder structure refactor

### LINKED ISSUE #19: Feature: Add operation status indicator
Once an operation has been applied, it's be great if we could add an icon that indicated that the action is in progress and then completed. This extends the existing issue #15.

I was thinking we could add the following unicode/non-unicode icons next to the resource names:
Deleting = ☠️ / -
Updating = ⚙️ / ~
Creating (if supported) = 🧱 / +

The end of the operation could be indicated by disappearance of the status icon.

Something similar to below:
icons

lawrencegripper:
Closes #19


lawrencegripper:
This is starting to look very possible. One downside is that the information is lagging by ~1min which makes me a bit unsure on how prominent to make this in the UI. 

`As a Public Preview service, the goal is currently that 95% of resource updates are available via Resource Graph within 1 minute.`

See: https://github.com/MicrosoftDocs/azure-docs/issues/19879

lawrencegripper:
screenshot from 2019-02-07 14-29-13

stuartleeks:
Do you know what the throttling limits are? This page refers to throttling but doesn't give any details: https://docs.microsoft.com/en-us/azure/governance/resource-graph/overview#throttling

lawrencegripper:
@stuartleeks good call, I've raised a feedback request here to see if they can be shared. 

https://github.com/MicrosoftDocs/azure-docs/issues/24478

I think I probably need to handle a graceful degradation when the graph isn't available or fails.

PR #49, opened 2019-02-19
Swagger codegen

stuartleeks:
Initial work to support swagger codegen to enable richer drill-down into resources/properties

This PR has `swagger-codegen` to generate code to create data structures representing the resource structure for azbrowse

Currently this supports `GET`, `DELETE` and `PUT` requests. It additionally supports logical `GET`s as AppService has a few requests that should have been `GET`s but are actuially `POST`s - this adds some slight complexity but allows you to easily drill into appsettings etc

`GET` requests are used for browsing, `DELETE` requests are handled by the `Del` existing key binding. `PUT`requests are handled by a new `Ctrl+U` key binding that opens the JSON in VS code for editing and then submits the changes - this makes is really easy to drill down into a resource (e.g. site/config/appsettings in AppService), edit and apply changes

Future work (for subsequent PRs) includes 
 * adding support for handling other verbs as actions, ideally parsing the body definition to help with creating new resources etc
* simplifying or automating adding and updating provider specs


stuartleeks:
@lawrencegripper - there are sstill a few rough edges around this, but I'd love to get your thoughts :-)

lawrencegripper:
Sure I've got a few things I'm trying to finish off today on another project but if I get some time I'll get stuck in. If I run out of time is Monday ok?

stuartleeks:
Of course! If you're really lucky I may get some refactoring done before then 😉

cmd/swagger-codegen/main.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+		if path.Operations.Get.Permitted {
+			fmt.Fprintf(w, "\t%s\tResourceType{\n", prefix)
+			getEndpoint := path.Operations.Get.Endpoint
+			if getEndpoint == nil {
+				getEndpoint = path.Endpoint
+			}
+			fmt.Fprintf(w, "\t%s\t\tDisplay:  \"%s\",\n", prefix, path.Name)
+
+			fmt.Fprintf(w, "\t%s\t\tEndpoint: mustGetEndpointInfoFromURL(\"%s\", \"%s\"),\n", prefix, getEndpoint.TemplateURL, getEndpoint.APIVersion)
```
lawrencegripper: [Nit] Would this be more readable using go templates? Traefik uses them when outputting stuff, helm is based on them and I know the AKS Engine team used to use them. I guess the pro is that the output is more visual when editing. [This is generating a TOML doc using them](https://github.com/containous/traefik-extra-service-fabric/blob/master/servicefabric_tmpl.go) https://golang.org/pkg/text/template/

Makefile
```diff
@@ -20,4 +20,9 @@ checks:
 	gometalinter --vendor --disable-all --enable=vet --enable=gofmt --enable=golint --enable=deadcode --enable=varcheck --enable=structcheck --enable=misspell --deadline=15m ./...
 
 ci-docker:
-	docker run -it -e BUILD_NUMBER=${TRAVIS_BUILD_NUMBER} -v $(CURDIR):/go/src/github.com/lawrencegripper/azbrowse golang:1.10 bash -f /go/src/github.com/lawrencegripper/azbrowse/scripts/release.sh
\ No newline at end of file
+	docker run -it -e BUILD_NUMBER=${TRAVIS_BUILD_NUMBER} -v $(CURDIR):/go/src/github.com/lawrencegripper/azbrowse golang:1.10 bash -f /go/src/github.com/lawrencegripper/azbrowse/scripts/release.sh
+
+swagger-codegen:
+	go run cmd/swagger-codegen/main.go --output-file ./internal/pkg/handlers/swagger.generated.go 
+	gofmt -s -w internal/pkg/handlers/swagger.generated.go
```
lawrencegripper: [nit] I think it would be good to add `go build internal/pkg/handlers/swagger.generated.go internal/pkg/handlers/types.go` here to check the output is building and isolate gen errors from other build errors. 

To make this line work I've pulled out `Register` into `Register.go` so we don't build all the other handlers too.

lawrencegripper: Do you think we could also add a test which loads `swagger.generated.go` and checks it doesn't panic? I guess the aim would be to automate the check that one of the paths isn't parsable by `mustGetEndpointInfoFromURL`

lawrencegripper: Happy to pick both of these up btw

internal/pkg/handlers/swagger.go
```diff
@@ -0,0 +1,190 @@
...
+	matchResult := resourceType.Endpoint.Match(currentItem.ExpandURL) // TODO - return the matches from getHandledTypeForURL to avoid re-calculating!
+	templateValues := matchResult.Values
+
+	if len(resourceType.SubResources) > 0 {
+		// We have defined subResources - Unmarshal the ARM response and add these to newItems
+		var resourceResponse armclient.ResourceResponse
+		err = json.Unmarshal([]byte(data), &resourceResponse)
+		if err != nil {
+			err = fmt.Errorf("Error unmarshalling response: %s\nURL:%s", err, currentItem.ExpandURL)
+			panic(err)
```
lawrencegripper: [future] I'm trying to move away from panic where possible. In cases like this would it be good to return an `expanderResult` with the error set. However, this is all over the codebase so I think this is something I'll pick up later and fix in one focused PR

cmd/azbrowse/main.go
```diff
@@ -341,6 +346,106 @@ func main() {
...
+		}
+
+		err = tmpFile.Close()
+		if err != nil {
+			status.Status(fmt.Sprintf("Cannot close temporary file: %s", err), false)
+			return err
+		}
+
+		newFilename := tmpFile.Name() + ".json"
+		err = os.Rename(tmpFile.Name(), newFilename)
```
lawrencegripper: You can use `*` as a wildcard in `ioutil.TempFile` like https://play.golang.org/p/gSZPfMM8YPV which hopefully removes the need for the rename unless it's used for anther reason

cmd/azbrowse/main.go
```diff
@@ -341,6 +346,106 @@ func main() {
...
+			status.Status(fmt.Sprintf("Cannot open edited file: %s", err), false)
+			return nil
+		}
+
+		updatedJSON := string(updatedJSONBytes)
+		if updatedJSON == originalJSON {
+			status.Status("No changes to JSON - no further action.", false)
+			return nil
+		}
+		if updatedJSON == "" {
```
lawrencegripper: [tweak] We could check that the JSON can be parsed with `json.UnMarshal` before submitting.

cmd/azbrowse/main.go
```diff
@@ -380,6 +485,32 @@ func main() {
...
+	}
+	if response["error"] != nil {
+		serializedError, err := json.Marshal(response["error"])
+		if err != nil {
+			err = fmt.Errorf("Error serializing error message: %s: %s", responseString, err)
+			return "", err
+		}
+		message := string(serializedError)
+		message = strings.Replace(message, "\r", "", -1)
+		message = strings.Replace(message, "\n", "", -1)
```
lawrencegripper: Nice tweak for readability I guess?

cmd/azbrowse/main.go
stuartleeks: Good idea - I'd been using invalid JSON as a way to test the error handling, but it makes sense to sanity check before sending :-)

cmd/azbrowse/main.go
stuartleeks: Yeah - the status section in the UI is too small to show the error when it contains line-breaks!

internal/pkg/handlers/swagger.go
stuartleeks: Good spot. I'd replicated the panics in the code I added but have been gradually migrating my code away from them - missed this one!

cmd/swagger-codegen/main.go
stuartleeks: Would you want this done before merging, or as a separate PR?

Makefile
stuartleeks: I'm happy for you to pick these up - sounds like you have a clear vision of what you want :-)

lawrencegripper: Cool, done :)

cmd/swagger-codegen/main.go
lawrencegripper: Yeah sounds good I've tracked it with #64

lawrencegripper:
Think this is good to go. I added some checks to to CI build to run `make swagger-generate` and check that it didn't cause any changes to ensure that PR's have committed the latest generated code and update the help to list CTRL+U command. 

My temp file fix requires go 1.12 so I've updated to use that in CI too.

Issue #24, opened 2019-01-03
Feature: Support parsing 'Deployment' Objects for nested browsing


lawrencegripper:
Currently opening 'Deployments' just shows the json array, it would be good to correctly parse these into a list that can be browsed easily.

PR #51, opened 2019-02-20
Support for browsing deployments #24 Bonus: Added color to the json

lawrencegripper:
image


lawrencegripper:
@stuartleeks  and @jjcollinge I'm going to merge this into `dev` so you can pick up latest for your branches.

Sound good?

Issue #18, opened 2018-12-21
Feature: Support rebinding keys from a config file


jjcollinge:
The current key bindings conflict with my tmux bindings - I'm sure this will be a common issue for people with various terminal setups. Being able to configure the bindings from a config file will let people have a custom setup.

lawrencegripper:
This is all done. Awesome work @jjcollinge

PR #52, opened 2019-02-20
Support custom key bindings

jjcollinge:
It's just a start - to demonstrate a possible way to support this.

This PR allows you to add a `bindings.json` file next to the binary of the format:
```json
{
    "ListActions": "Ctrl+G",
    "ListRefresh": "F8"
}
```
This will override the default command for those key handlers.

Todo:
- [X] Support custom bindings
- [ ] Testing
- [X] Readme
- [X] Dynamic views

Issue: #18


lawrencegripper:
Nice, I like the idea of pulling out the views into their own package.

jjcollinge:
@lawrencegripper sure you're going to love how I handled those conflicting keys...

internal/pkg/keybindings/mappings.go
```diff
@@ -0,0 +1,133 @@
...
+	"f4":         gocui.KeyF4,
+	"f5":         gocui.KeyF5,
+	"f6":         gocui.KeyF6,
+	"f7":         gocui.KeyF7,
+	"f8":         gocui.KeyF8,
+	"f9":         gocui.KeyF9,
+	"f10":        gocui.KeyF10,
+	"f11":        gocui.KeyF11,
+	"f12":        gocui.KeyF12,
+	"enter":      gocui.KeyEnter,
```
lawrencegripper: This is some serious dedication!

internal/pkg/keybindings/mappings.go
```diff
@@ -0,0 +1,133 @@
...
+	gocui.KeyCtrlRsqBracket: "ctrl+] or ctrl+5",
+	gocui.KeyCtrlUnderscore: "ctrl+_ or ctrl+7",
+	gocui.KeyCtrlA:          "ctrl+a",
+	gocui.KeyCtrlB:          "ctrl+b",
+	gocui.KeyCtrlC:          "ctrl+c",
+	gocui.KeyCtrlD:          "ctrl+d",
+	gocui.KeyCtrlE:          "ctrl+e",
+	gocui.KeyCtrlF:          "ctrl+f",
+	gocui.KeyCtrlG:          "ctrl+g",
+	gocui.KeyCtrlH:          "ctrl+h or ctrl+backspace",
```
lawrencegripper: Gotta love a bit of legacy

internal/pkg/views/treeview.go
```diff
@@ -228,7 +231,7 @@ func (w *ListWidget) ExpandCurrentSelection() {
 				}
 				// Log that we have a primary response
 				hasPrimaryResponse = true
-				w.contentView.SetContent(done.Response, "[CTRL+F -> Fullscreen|CTRL+A -> Actions] "+currentItem.Name)
+				w.contentView.SetContent(done.Response, fmt.Sprintf("[%s-> Fullscreen|%s -> Actions] %s", strings.ToUpper(w.FullscreenKeyBinding), strings.ToUpper(w.ActionKeyBinding), currentItem.Name))
```
lawrencegripper: Very nice touch!

jjcollinge:
@lawrencegripper could you test and review this please. I've not written any tests yet but I'll pick that up as a different issue that I'll do on my flight on Sunday - it'll save me keep fixing conflicts :P - think I merged in ok but worth double checking.

jjcollinge:
Added dynamic padding for the help view to save any issues with all that whitespace

martinpeck:
@jjcollinge Out of interest, do any of these changes prevent `TAB` being confused with `Ctrl+I` (to open help)?

I found myself wanting to hit `TAB` to jump back/forth from the list to the details. However, `TAB` and `Ctrl+I` are often mapped to the same thing.

So, will this make it easier to map the Help page to something else, and allow jumping back and forth using `TAB`?

lawrencegripper:
@jjcollinge I think this looks ace. I've pushed a quick update which guards against the same key being bound to multiple handlers and gives a nice error message to the user if they do something like that. 

```
The key `up` is already in use by `listup`. Please update your `~/.azbrowse-bindings.json` file to a valid configuration and restart
```

Hope you don't mind me pushing this to the branch.

PR #58, opened 2019-02-21
Make the Sun Shine on MacOS and Fix #55

### LINKED ISSUE #55: Resource status font does not display correctly on MacOS
I've tried various fonts, and I've tried iTerm2 vs Terminal, but I still get a question mark instead of the expected unicode glyph when I build and run `azbrowse`. I've not investigated this, and fully intend do, but wanted to log this before I forget.

image

martinpeck:
Fixes https://github.com/lawrencegripper/azbrowse/issues/55 by replacing the current "white sun" glyph...

https://unicode-table.com/en/1F323/

... with the "white sun with rays" glyph...

https://unicode-table.com/en/263C/

... as the former appears to be unsupported on MacOS

### DIFF
--- a/cmd/azbrowse/help.go
+++ b/cmd/azbrowse/help.go
@@ -17,42 +17,42 @@ func ToggleHelpView(g *gocui.Gui) {
 func DrawHelp(v *gocui.View) {
 
 	fmt.Fprint(v, style.Header(`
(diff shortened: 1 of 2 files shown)


martinpeck:
Looks like this...

image

jjcollinge:
The current one is also unsupported on my linux setup

martinpeck:
Doh! Will do. Stupid editor cleaning things up!

martinpeck:
@lawrencegripper the whitespace is restored, leaving the icon change as the only change in that file. Sorry about that!

PR #63, opened 2019-02-26
Support Access Token Caching

martinpeck:
In the current codebase, the command `az account get-access-token` is invoked for every `armclient` request. This is fine, but on some basic testing I see approx 500ms overhead for this call (I assume part of that is caused by having to spawn the sub process).

I've started to experiment with caching that token. At first, I'm simply caching it indefinitely, which will cause problems when the token expires. I changed the token structure so that I can grab the ExpiresOn value from the json response, however, that ExpiresOn looks like this...

```json
"expiresOn": "2019-02-26 01:56:50.210663"
```

It doesn't appear to include a timezone. That makes it hard to parse. In addition, when I run the `az account get-access-token` command in some shells (e.g. the `shell.azure.com` portal) the `expiresOn` value is `N/A`. So, I'm not sure I can reliably use this value.

According to `az account get-access-token --help`:

> The token will be valid for at least 5 minutes with the maximum at 60 minutes.

So, one option is the simply cache the token for 5 mins and re-request it after that. Anyone using the tool for > 5 mins would see an additional 500ms.


martinpeck:
One other thing I discovered, while digging around here, is that the Azure Functions Core Tools uses this "shell out to az" approach too, but they have additional code that tries various strategies to get the token, so that the code works on other platforms and cmd/PowerShell. 

e.g. https://github.com/Azure/azure-functions-core-tools/blob/dev/src/Azure.Functions.Cli/Actions/AzureActions/BaseAzureAction.cs#L92

Is this something that needs to happen here?

martinpeck:
The other approach that might be helpful here is to directly access `$HOME/.azure/accessTokens.json`, where these tokens are written by `az`.

lawrencegripper:
The consensus seems to be that accessing the `azureTokens.json` directly isn't a good idea. See: https://github.com/terraform-providers/terraform-provider-azurerm/pull/1752

lawrencegripper:
Firstly this is awesome and great idea!

Maybe a nice middle ground would be to cache the token and then intercept any `403` requests which comeback from the `armclient` when making a call. At this point the code could then request a new token from `az` and retry the request. 

That way we wouldn't have an issue with clock drift/timezones etc and support those who have tokens with a longer `TTL` maybe at the expense of the odd retried call every 5mins. 

What do you think? I've not thought it through fully so maybe so big holes in this plan.

lawrencegripper:
Regarding the expires issue with `shell.azure.com` that occurs because the token is actually a Managed Service Identity. 

The Terraform AzureRM codebase handles this case and is written in go so might be an interesting bit to look at (although the use case is different so may not suit perfectly). https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/config.go#L391

martinpeck:
Yeah, I wasn't keen on hacking away in the accessTokens.json file. It would have been nice is `az` returned all of the data (e.g. refresh tokens) but never mind. Reacting to a 403 might well be a good way to deal with this too.

jjcollinge:
How would you differentiate between an expired 403 and a valid 403 due to a bad token? Handling no token would be easy enough but is there a scenario where a token may exist but be invalid? You could end up in a recursive loop

martinpeck:
Plan is to retry once. If there’s a 403 then try again, but only once.

lawrencegripper:
That makes sense to me

lawrencegripper:
Would you be offended if I picked this one up based off your work so far?

martinpeck:
@lawrencegripper sure...go for it.

I’m really sorry that I’ve not had time to see this through

stuartleeks:
Another option to get the token expiry would be to decode the JWT and use the `exp` field

lawrencegripper:
Closing as solved by #89 - without this work I wouldn't have even thought of this as a source of perf problems so massive thanks!

PR #71, opened 2019-03-03
Update README.md

scotty-c:
A small fix for the readme to mount local time into the container for people outside of PST. Otherwise you an error your token is invalid as it is before the container date. For example, I am in Australia and the token thinks it is in the future.


lawrencegripper:
@scotty-c awesome catch, thanks!

scotty-c:
No worries, happy to help. Great project !!!

PR #74, opened 2019-03-06
Script update process for swagger specs

### LINKED ISSUE #64: Code-gen: Move to using Go template language for generated code.
Currently the `swagger-generator` uses a writer. To improve readability we could look at moving to using a `GoTemplate` 

https://github.com/lawrencegripper/azbrowse/pull/49#discussion-diff-260254251R317

stuartleeks:
Add `update-swagger.sh` to pull latest swagger specs into azbrowse repo in the structure that `swagger-codegen` expects (previously it was a manual process)
Update `swagger-codegen` to use https://github.com/go-openapi/ to parse swagger specs rather than extend the previous home-grown parsing to accommodate differences across the full API spec set
Update specs and generated code
Update the codegen to use go templates (fixes #64)


stuartleeks:
hold fire - I've just realised that there's some refactoring I want to do :-)

stuartleeks:
Refactoring done - also converted to go templates for the codegen :-)
The PR is quite big now, so let me know if you'd like me to rebase and split it into separate PRs (even just taking out the updates of the swagger specs and generated code into a separate PR could make it easier to review!)

lawrencegripper:
I've update the PR to run the `swagger-update` process as part of the CI process to ensure the swagger committed is generated from the process not manually. (If it is then build will pass, if it makes changes to the files it will fail)

PR #77, opened 2019-03-12
Add deployment operations under deployments

stuartleeks:
Add new DeploymentExpander to show operations under deployments


internal/pkg/handlers/activitylog.go
```diff
@@ -61,7 +61,7 @@ func (e *ActivityLogExpander) Expand(ctx context.Context, currentItem *TreeNode)
 			ID:              log.ID,
 			Parentid:        currentItem.ID,
 			ExpandURL:       ExpandURLNotSupported,
-			ItemType:        subDeploymentType,
```
lawrencegripper: Woops! Sorry.

internal/pkg/handlers/deployment.go
```diff
@@ -0,0 +1,78 @@
+package handlers
```
lawrencegripper: nit: To keep stuff tidy and as they've quite coupled I think it would be best to sit this in the existing `deployments.go` file. We've done this in other areas of the codebase and it's quite commmon in go to group related structs into files. What do you think?

stuartleeks: Done. (I actually started that way and then changed it!)

stuartleeks:
Latest commits revert the display for deployment JSON, and add ResourceType/Name to the tree node

stuartleeks:
Rebased on latest master changes :-)

stuartleeks:
Looks like the build failure is independent of these changes. Have created #80 to address :-)

lawrencegripper:
Awesome looks good, I've made one person preference tweak which is to use the `ResourceName` field as the title when it's available instead of the `OperationId` so it's more in-line with the portal and helps highlight key data. 

topleveldetails

image

PR #85, opened 2019-04-01
Working basics of multi-delete #71

lawrencegripper:
pendingDelete

Oustanding work:
- [x] Prevent same item being added more than once 
- [x] Show errors in delete process (currently swallowed)
- [x] Look at moving to the ARM batch API to submit all deletes as one request.


lawrencegripper:
From some testing I'm happy with this as being MUCH better than the existing behavior so going to merge and look to get feedback.

internal/pkg/views/notifications.go
```diff
@@ -0,0 +1,175 @@
...
+			Message: "Item `" + display + "` doesn't support delete",
+			Timeout: time.Second * 5,
+		})
+		return
+	}
+	gui.Update(func(g *gocui.Gui) error {
+		deleteMutex.Lock()
+		defer deleteMutex.Unlock()
+
+		for _, i := range pendingDeletes {
```
jjcollinge: I know we talked about this but I've forgotten why this needed to be an ordered list rather than a map?

internal/pkg/views/notifications.go
```diff
@@ -0,0 +1,175 @@
...
+		defer deleteMutex.Unlock()
+
+		pending := pendingDeletes
+		event, _ := eventing.SendStatusEvent(eventing.StatusEvent{
+			InProgress: true,
+			Message:    "Starting to delete items",
+			Timeout:    time.Second * 15,
+		})
+
+		for _, i := range pending {
```
jjcollinge: Obviously not performance optimal but probably gives the most intuiative user experience i.e. sequential deletes and notifications for failures.

internal/pkg/views/notifications.go
```diff
@@ -0,0 +1,175 @@
...
+			_, err := armclient.DoRequest(context.Background(), "DELETE", i.url)
+			if err != nil {
+				event.Failure = true
+				event.Message = "Failed to delete `" + i.display + "` with error:" + err.Error()
+				event.Update()
+
+				// In the event that a delete fails in the
+				// batch of pending deletes lets give up on the rest
+				// as something might have gone wrong and best
```
jjcollinge: Understandable. I'm in 2 minds about this. It could be dangerous to change this behaviour later if users get used to it and as it stands it implies the delete order is the same as the pending queue order which may not always be the case? i.e. if you move to batch or map. Likely not an issue as it'd probably be rare for a user to pick up on the subtleties but just thought I'd call it out.

internal/pkg/views/notifications.go
```diff
@@ -0,0 +1,175 @@
...
+	})
+}
+
+// ConfirmDelete delete all queued/pending deletes
+func ConfirmDelete() {
+	gui.Update(func(g *gocui.Gui) error {
+		deleteMutex.Lock()
+		defer deleteMutex.Unlock()
+
+		pending := pendingDeletes
```
jjcollinge: In my testing I did not receive this notification and it simply forze for ~30 seconds and then said deletion complete.

internal/pkg/views/notifications.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Layout draws the widget in the gocui view
+func (w *NotificationWidget) Layout(g *gocui.Gui) error {
+	// Don't draw anything if no pending deletes
+	if len(pendingDeletes) < 1 {
+		g.DeleteView(w.name)
+		return nil
+	}
+
+	height := len(pendingDeletes)*1 + 7
```
jjcollinge: Does this need to be capped with a scroll or something?

internal/pkg/views/notifications.go
```diff
@@ -0,0 +1,175 @@
...
+	pending := pendingDeletes
+
+	fmt.Fprintln(v, style.Title("Pending Deletes:"))
+	for _, i := range pending {
+		fmt.Fprintln(v, " - "+i.display)
+	}
+	fmt.Fprintln(v, "")
+	fmt.Fprintln(v, "Do you want to delete these items?")
+	fmt.Fprintln(v, style.Warning("Press "+strings.ToUpper(w.ConfirmDeleteKeyBinding)+" to DELETE"))
+	fmt.Fprintln(v, style.Highlight("Press "+strings.ToUpper(w.ClearPendingDeletesKeyBinding)+" to CANCEL"))
```
jjcollinge: This looks really good and works well!

internal/pkg/views/notifications.go
lawrencegripper: I think it's mainly because I feel it's more intuitive for the list to be ordered so you add something and it goes on the bottom.

Not super computer science wise but hopefully the code is easy to follow - what do you think?

internal/pkg/views/notifications.go
lawrencegripper: Yeah I also kinda think it's not the type of action I'm going to be super worried about going instantly, if anything the slightly slower nature lets you bail out with a `CTRL+C` if you realise you messed up the list.

internal/pkg/views/notifications.go
lawrencegripper: Yeah good point. Having looked more at the Batch API it can't handle more than 20 requests at a time so feels like there would always be some form of paging going on. 

My gut at the moment is to give up on the Batch API and leave this behavior as is - in my personal use case I'm usually only deleting 3/4 things at a time but I may not be normal.

internal/pkg/views/notifications.go
lawrencegripper: Ah nice great spot - looks like I've messed up and somehow the requests are happening on the UI thread (need to dig into GOCUI a bit more and work out how it works). 

I've fixed this up by moving out of `Update` call and moving to a go routine then relying on the `SendStatusEvent` to trigger updates on the UI which seems to work nicely.

internal/pkg/views/notifications.go
lawrencegripper: Another awesome spot. Thinking about it I'm not keep on scrolling as could easily miss items that are at the top of the list and think they're not on delete list. What do you think of something like this inside `AddPendingDelete`?

```golang

	// Don't add more items than we can draw on the
	// current terminal size
	_, yMax := gui.Size()
	if len(pendingDeletes) > (yMax - 12) {
		eventing.SendStatusEvent(eventing.StatusEvent{
			Failure: true,
			Message: "Can't add `" + display + "` run out of space!",
			Timeout: time.Second * 5,
		})
		return
	}
```

jjcollinge: Seems fair enough yh

lawrencegripper:
I've added some more testing to this and caught a couple of extra bugs which are now fixed. 

1. If you CTRL+Y and then attempt to add more items to the list they are added during the delete and then cleared unexpectedly at the end
2. `armclient` had a bug which prevented it from identifying errors in responses
3. The locking structure was a bit off to prevent re-entry into the methods. I've tweaked this to be safer (hopefully)

To enable to better testing I've moved the global `AddPendingDelete` etc methods to the struct (tests would bleed into one another causing confusion). 

All in all I think this is hopefully in a good state to be merged. One problem is testing the new integration tests in CircleCI as they need `tty` to work as expected which looks like it doesn't. I'm playing with launching an instance of `xterm` to run the tests and then somehow getting the outcome from the script but still fiddling.

lawrencegripper:
Woooooo! Now have integration tests running under Azure Devops with X server, Dummy display driver, XTerm and some voodoo :)

Issue #86, opened 2019-04-06
Migrate from dep to go modules


jjcollinge:
I think it probably makes sense to migrate from deps to go modules as that's the way forward for the language and is in a reasonably stable state. I'll take a first look at this today as a way of getting familiar with go modules.

lawrencegripper:
Done! Awesome work @jjcollinge!

PR #87, opened 2019-04-06
Migrate deps to go mod

jjcollinge:
Issue #86


jjcollinge:
Sorry about the messy history - some reason I did the initial migrate on a stale branch which worked locally but failed in CI then had to fix. We can clean that up on merge if we decide to go ahead.

lawrencegripper:
Wow this is awesome - thank. Took me a while to read up on modules stuff to be able to understand whats going on. 

One of the things I'm a bit worried about with the move, likely me being paranoid, is what happens in the `go modules` world if one of the projects dependencies github repos is removed? As Github isn't a package manager feels like this a bit of a risk. Looks like I can help my paranoia by [running this command](https://tip.golang.org/cmd/go/#hdr-Modules_and_vendoring) but yeah not sure whether it's me being overly cautious or not.

jjcollinge:
Yh I think this is something project Athens was trying to address. Not sure how far it has progressed, will have a look.

lawrencegripper:
Nice that sounds like it sovles the problem nicely, fingers crossed it works!

Issue #96, opened 2019-06-26
Bug: Editing an item with a terminal editor stops azbrowse functioning correctly


stuartleeks:
If azbrowse is configured to use a terminal-based editor (e.g. vi, nano) then azbrowse stops functioning correctly when the editor exits.

Symptoms noticed so far
* screen is not fully refreshed, so parts of the editor are still present
* navigating the list doesn't update the selected item

Issue #91, opened 2019-05-25
Respect alternatives system for default editor


ohthehugemanatee:
Most Linux distros implement the [alternatives system](https://linux.die.net/man/8/update-alternatives) to let users configure default programs across the alternatives that are installed on their system. (choose your editor with `update-alternatives --config editor`).

We should respect this configuration.

lawrencegripper:
Love this idea, we've got to be a bit conscious that the tools runs over Mac, Linux and Windows so whatever we do should be cross plat. 

Would love to see a PR to tackle this one or proposal around how to tackle. 

One option which springs to mind is maybe using platform specific rules where appropriate and allowing an override in a settings file or envar. 

We currently use `[https://github.com/skratchdot/open-golang/](https://github.com/skratchdot/open-golang/)` to select the correct browser. It may already have the logic to select the correct editor for a `.json` file. 

One complexity is monitoring the editor we've loaded to see when it's `done` with the file. I think the current method is `vscode` specific and monitors the `tab` opened. One option is to monitor the process but for multi-tabbed editors this means closing the whole editor to continue, not ideal. Another would be to monitor the file and observe save operations but this might not be the UX people expect `Updated Azure on Save` feels like it might catch people out. 

What do you think?

cc @stuartleeks

stuartleeks:
I am looking at something related to this to enable Vs code to work as an editor on WSL so will take a look 😀

lawrencegripper:
This this is good to close now as it's shipped in #97

PR #97, opened 2019-06-26
Make the ListUpdate editor configurable

stuartleeks:
This PR makes the editor used for the ListUpdate action configurable.

The editor command and arguments can be configured in addition to controlling whether the temp file path should be converted from Linux to Windows paths under WSL. By default VS Code is still used as the editor and whether or not to translate the path is dynamically determined.

Note: currently this only supports graphical editors. Terminal-based editors still need more work - see #96

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -16,7 +16,7 @@
             "env": {},
             "args": [],
             // run azbrowse in a terminal with the following command before starting the debugger
-            // go build &&  dlv exec ./azbrowse --headless --listen localhost:2345 --api-version 2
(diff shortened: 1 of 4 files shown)


lawrencegripper:
Nice will try and have a look soon, tagging #91 and @ohthehugemanatee too

stuartleeks:
I haven't tried to tackle #91 yet - I think it makes sense to resolve #96 first. I've taken what I think is a step in the right direction and figured that it added some value against the previously hard-coded approach to open a PR

lawrencegripper:
Just re-queuing the build - not quite sure why the integration tests are failing for this one now.

PR #109, opened 2019-08-29
Support viewing container instance logs

lawrencegripper:
#104

### DIFF
--- a/pkg/armclient/types.go
+++ b/pkg/armclient/types.go
@@ -286,6 +286,11 @@ type ActivityLogResource struct {
 	} `json:"value"`
 }
 
+// ContainerLogResponse for container logs
+type ContainerLogResponse struct {
+	Content string `json:"content"`
+}
+
 // ContainerGroupResponse is the response to a get request on a container group
 type ContainerGroupResponse struct {
 	ID         string `json:"id"`
--- a/pkg/armclient/types.go
+++ b/pkg/armclient/types.go
@@ -285,3 +285,73 @@ type ActivityLogResource struct {
 		} `json:"httpRequest,omitempty"`
 	} `json:"value"`
 }
+
+// ContainerGroupResponse is the response to a get request on a container group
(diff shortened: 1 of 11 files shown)


stuartleeks:
Nice!!

I just tried it on an ACI instance with quite large logs and the perf of scrolling the list wasn't great. Looking at the [docs](https://docs.microsoft.com/en-us/rest/api/container-instances/container/listlogs) there is a `tail` parameter - I wonder if we should consider setting a value for that (without it we get the full logs up to 4MB!)

Also, it really makes me want page up/down support in the details view :-)

lawrencegripper:
@stuartleeks hopefully this now behaves a bit better, I've limited the logs to 400 lines and improved the formatting logic.

stuartleeks:
I don't have the previous ACI instance to test against but just created one with a reasonable log size and perf definitely seemed better :-)

lawrencegripper:
Nice, working with this a lot last night I CAN'T ship without pageup/pagedown as it's nighmareish to use. I'm calling time on this flakey test as it's doing my head in. 

Will try and get those fixes in then can look to merge.

lawrencegripper:
Think this is good to go now, anyone able to do a quick review for me?

pkg/armclient/types.go
```diff
@@ -285,3 +285,78 @@ type ActivityLogResource struct {
...
 	} `json:"value"`
 }
+
+// ContainerLogResponse for container logs
+type ContainerLogResponse struct {
+	Content string `json:"content"`
+}
+
+// ContainerGroupResponse is the response to a get request on a container group
+type ContainerGroupResponse struct {
```
stuartleeks: Do we want to put these types in here? Could become quite large as we add more data plane functionality.

lawrencegripper: Yeah I agree, I'll track an issue for re-orging this stuff in a separate PR.

PR #115, opened 2019-09-04
Initial work on displaying metrics

lawrencegripper:
The plan here is to enumerate the available metrics for each service and then graph the time series items in the display panel... currently WIP


lawrencegripper:
So it looks like this is actually doable!

Currently the PR will dynamically discover the metric namespace for the resource, query it's available metrics and build a query with the right aggregations then pull back the data and build an ascii graph with it....

Here is an example of the server-farm memory usage graph for last 3 hours. 

graphs

lawrencegripper:
metricstake2

Done some refactoring, pretty happy this is ready to ship!

PR #116, opened 2019-09-18
Add support for viewing App Insights saved queries /functions (aka Analytics Items)

stuartleeks:


internal/pkg/handlers/appInsights.go
```diff
@@ -0,0 +1,175 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/lawrencegripper/azbrowse/internal/pkg/style"
+	"github.com/lawrencegripper/azbrowse/pkg/armclient"
+)
+
+type analyticsItem struct {
+	Content      string `json:"Content"`
```
lawrencegripper: Nice - This is an interesting option for types that aren't going to be re-used by other handlers. Maybe a nice approach going forward, I could pull the ones I added for metrics into the handlers.

lawrencegripper:
Soooo just going to leave this here... We could add an "Execute" option with https://github.com/olekukonko/tablewriter

stuartleeks:
Yup, happy for this to be merged

stuartleeks:
And lol re the table writer.... I could actually use that for something else that I was thinking of :-)

PR #120, opened 2019-10-04
Add support for terminal editors

stuartleeks:
Add support for terminal editors (e.g. vim, nano)
Fix bug with default VS Code editing under WSL (due to changes in VS Code path mapping)


internal/pkg/keybindings/listhandlers.go
```diff
@@ -432,13 +435,12 @@ func (h ListUpdateHandler) getEditorConfig() (config.EditorConfig, error) {
...
 	}
 	// generate default config
-	translateFilePathForWSL := wsl.IsWSL() // If on WSL then translate path so that it is valid when loaded by VS code in Windows
 	return config.EditorConfig{
 		Command: config.CommandConfig{
 			Executable: "code",
 			Arguments:  []string{"--wait"},
 		},
-		TranslateFilePathForWSL: translateFilePathForWSL,
+		TranslateFilePathForWSL: false, // previously used wsl.IsWSL to determine whether to translate path, but VSCode  now performs translation from WSL (so we get a bad path if we have translated it)
```
lawrencegripper: nit: If we no longer need this can we remove the property from the `EditorConfig` struct?

Ah wait I've misunderstood this is only when using VSCode... gotcha

internal/pkg/views/treeview.go
```diff
@@ -6,11 +6,12 @@ import (
 	"strings"
 	"time"
 
-	"github.com/jroimartin/gocui"
 	"github.com/lawrencegripper/azbrowse/internal/pkg/eventing"
 	"github.com/lawrencegripper/azbrowse/internal/pkg/handlers"
 	"github.com/lawrencegripper/azbrowse/internal/pkg/style"
 	"github.com/lawrencegripper/azbrowse/internal/pkg/tracing"
+	"github.com/stuartleeks/gocui"
+	// "github.com/nsf/termbox-go"
```
lawrencegripper: nit: Remove commented out code

internal/pkg/views/treeview.go
```diff
@@ -142,6 +143,9 @@ func (w *ListWidget) ExpandCurrentSelection() {
 		w.title = ""
 	}
 
+	// termbox.Sync()
+	// termbox.Flush()
+	// termbox.Clear(0,0)
```
lawrencegripper: nit: Remove commented out code

internal/pkg/keybindings/listhandlers.go
```diff
@@ -432,13 +435,12 @@ func (h ListUpdateHandler) getEditorConfig() (config.EditorConfig, error) {
 		return userConfig.Editor, nil
```
lawrencegripper: future: One odd thing I hit was when I set "vim" but didn't have "vim" installed only "vi" the edit panel will hang. Maybe worth pulling out an issue to do some validation on the executable being in the path and throwing a more human error if it's not but it's not a blocker to this work for me.

internal/pkg/keybindings/listhandlers.go
stuartleeks: Yup, still needed in config :-)

internal/pkg/views/treeview.go
stuartleeks: Hmm, thought I had removed those. Weird!

internal/pkg/keybindings/listhandlers.go
stuartleeks: Ah, good spot - I wasn't reinstating the alt buffer if there is an error launching the editor!

Have added a commit to fix this

PR #121, opened 2019-10-07
Add navigate cmdline arg

stuartleeks:
New arg allows navigation to a specific resource, e.g. 

`azbrowse navigate /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/myGroup/providers/Microsoft.Web/site/mysite`


cmd/azbrowse/main.go
```diff
@@ -212,20 +238,69 @@ func main() {
...
 
 	}()
 
+	if navigateToID != "" {
+		go func() {
+			navigatedChannel := eventing.SubscribeToTopic("list.navigated")
+			var lastNavigatedNode *handlers.TreeNode
+
+			for {
+				nodeListInterface := <-navigatedChannel
```
lawrencegripper: Not sure if this is a problem, going to write it down and see. 

If you start with this method but an expander times out then this go routine will carry on running .... listening for `list.navigated` which will be generated if you keep running after the failure and moving around. This could then result in some jumping around later as you landed on the item that starts with the id...

hmm I think this is a real edge case. Maybe a fix would be to subscribe to `statusEvents` and exit the loop if one is generated with a failure but this isn't ideal as a non-essantial handler could have generated the error. 

Maybe simplest is to from from `for` -> `select` with a timeout set so this go routine exists after 45secs if it hasn't got to the item?

Also valid to ignore this.

PR #124, opened 2019-10-08
Ideas for filter

stuartleeks:
A few ideas for filtering:

* Highlight the filter matches
* Use the Back action to clear the filter if set. Once cleared, normal Back action takes effect
* Use Display text for filtering rather than name as Display is what the user sees
* Trigger redraw so that the applied filter matches what user has typed
* Add the filter text to the list title (persists after command palette is dismissed)
* Add a command to clear the filter (defaults to Esc which feels natural to me)


stuartleeks:
@lawrencegripper with the filter highlighting, it looks like the filtering is a character behind the filter text that the user has typed!

lawrencegripper:
These are AWESOME. Pulling into my PR now :)

Issue #21, opened 2018-12-21
Feature: Get resource URL


jjcollinge:
I think it'd be useful to have an option to just grab the URL of the currently selected resource. This could be used if you want to share a resource, save it for other uses or browse it in a non default browser.

PR #128, opened 2019-10-11
Integrate with Kubernetes API

stuartleeks:
Initial WIP for drilling into the Kubernetes API for AKS clusters

TODO list:
 - [X] get cluster credentials
 - [X] get cluster swagger
 - [X] create nodes from swagger dynamically
 - [x] reuse swagger parsing when revisiting root k8s api node
 - [ ] testing!
 - [x] Remove duplicate nodes under `/api/v1`
 - [x] handle deletion
 - [X] refactoring/renaming classes now that they are better understood
 - [X] colorise YAML
 - [ ] Handle updates for non ARM resources
 - [x] ~~stretch/future PR:~~ handle logs requiring extra parameter for container name
 - [~~stretch/future PR: add command to copy node IDs (e.g. for use in `navigate` command) UPDATE - this is #21~~


stuartleeks:
I can't decide whether to add code to avoid requerying the swagger etc for a cluster. Perf doesn't seem too bad, and requerying means that you would pick up any new CRDs which could be a benefit. Any thoughts @lawrencegripper?

stuartleeks:
Cool, I'll try to take a look at the cache stuff.

Let me know how your general testing goes

stuartleeks:
OK- latest commit has caching of the k8s swagger etc

lawrencegripper:
WOW - I want this merged so badly!

image

image

internal/pkg/handlers/containerService-swaggerApiSet.go
```diff
@@ -0,0 +1,208 @@
...
+		data := string(buf)
+		return data, nil
+	}
+	return "", fmt.Errorf("Response failed with %s (%s)", response.Status, url)
+}
+
+// ExpandResource returns metadata about child resources of the specified resource node
+func (c SwaggerAPISetContainerService) ExpandResource(ctx context.Context, currentItem *TreeNode, resourceType swagger.ResourceType) (APISetExpandResponse, error) {
+
+	if resourceType.Endpoint.TemplateURL == "/api/v1/namespaces/{namespace}/pods/{name}/log" {
```
lawrencegripper: Very cool that you can get logs :)

internal/pkg/handlers/containerService.go
```diff
@@ -0,0 +1,330 @@
...
+	return false, nil
+}
+
+// Expand returns ManagementPolicies in the StorageAccount
+func (e *AzureKubernetesServiceExpander) Expand(ctx context.Context, currentItem *TreeNode) ExpanderResult {
+
+	swaggerResourceType := currentItem.SwaggerResourceType
+	if currentItem.Namespace != "AzureKubernetesService" &&
+		swaggerResourceType != nil &&
+		swaggerResourceType.Endpoint.TemplateURL == "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.ContainerService/managedClusters/{resourceName}" {
```
lawrencegripper: nit: maybe pull out into a const

lawrencegripper: Changed my mind... think it'd be less clear as a const

stuartleeks: #done

lawrencegripper:
@stuartleeks you happy for this to merge?

stuartleeks:
yup, happy to merge - have some more improvements on it but happy to put them in a new PR

stuartleeks:
Need a review btw ;-)

internal/pkg/handlers/containerService.go
```diff
@@ -0,0 +1,332 @@
...
+		},
+		SuppressAPIVersion: true,
+	}
+	var paths []*swagger.Path
+	paths, err = swagger.MergeSwaggerDoc(paths, &config, doc, false)
+	if err != nil {
+		return swaggerResourceTypes, err
+	}
+
+	// uncomment this code to help debug the generated API paths!
```
jjcollinge: @lawrencegripper what's the wider story about enabling debug features - do we drive all this behaviour via the `--debug` flag? Do we need to look for build macro alternatives or do you think uncomment and rebuild is sufficient?

internal/pkg/handlers/containerService.go
stuartleeks: It is repeated in a couple of places, so I opted for const :-)

internal/pkg/handlers/containerService.go
stuartleeks: Just pushed an update that outputs this based on whether debug is set

stuartleeks:
YAML colourisation just snuck in :-)

lawrencegripper:
I'm declaring this PR a MONSTER!

PR #138, opened 2019-10-21
Remove Microsoft.Web workaround

### LINKED ISSUE #137: Fix CI build - Microsoft.Web specs updated
The CI build is failing as per https://circleci.com/gh/lawrencegripper/azbrowse/364.

This is because there was a workaround in the swagger update script to handle a partial spec folder for Microsoft.Web. That team has now released newer specs with a complete folder so the workaround is no longer needed (and intentionally failed to require an update)

Workaround script:

```bash
# Temporary fixup for Microsoft.Web
# The latest version is 2018-02-01, except for Certificates.json which has 2018-11-01
# Since the 2018-11-01 Certificates has no real difference (and has references to the 2018-02-01 folder)
# manually copy back the 2018-02-01 files
echo "WARNING: Forcing 2018-02-01 version of web/Microsoft.Web as a temporary fix"
rm swagger-specs/top-level/web/Microsoft.Web/stable/2018-11-01 -r # this will fail if there is a newer version than 2018-11-01, which is a good sanity check
mkdir swagger-specs/top-level/web/Microsoft.Web/stable/2018-02-01 --parents
cp $ApiRepo/specification/web/resource-manager/Microsoft.Web/stable/2018-02-01/*.json swagger-specs/top-level/web/Microsoft.Web/stable/2018-02-01

```

stuartleeks:
Fixes #137


lawrencegripper:
Quick question: Wouldn't we expect the CircleCI build to now be passing for this one with the spec updated and workaround removed?

stuartleeks:
@lawrencegripper - That was my expectation! See #140 for my latest thoughts :-)

PR #146, opened 2019-11-04
Add basic support for Azure Search

stuartleeks:
Extend tests for EndpointInfo and add support for parsing/building URLS such as `/indexes('{indexName}')
Refactor swagger-codegen to make it easier to include data-plane codegen (note that this has re-generated the management plane swagger and picked up new API versions)
Add Azure search expander

Rebased on top of #145


stuartleeks:
@lawrencegripper  - I'll rebase this on your refactoring when it's complete

PR #152, opened 2019-11-07
Adding expander tests

lawrencegripper:
This adds a set of integration tests for the core set of expanders

- [x] Tenant Epxander
- [x] Subscription Expander
- [ ] Resource Group Expander


internal/pkg/expanders/integration_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"gopkg.in/h2non/gock.v1"
+)
+
+func Test_Expanders(t *testing.T) {
+
+	InitializeExpanders(nil)
+	expanders := getRegisteredExpanders()
+
+	// Add special case of default expander
+	expanders = append(expanders, &DefaultExpander{})
```
lawrencegripper: @stuartleeks and @jjcollinge what do you think of this approach?

lawrencegripper: (this section of file not this line in particular)

internal/pkg/expanders/testdata/armsamples/resource/failingResource.json
```diff
@@ -0,0 +1,14 @@
+{
+    "id": "/subscriptions/5774ad8f-0000-0000-0000-0447910568d3/resourceGroups/stable/providers/Microsoft.ContainerRegistry/registries/test",
```
jjcollinge: should these IDs be in?

lawrencegripper: Fixed thanks

PR #155, opened 2019-11-08
Add Azure Search query invocation

stuartleeks:
Add new command to allow search queries to be invoked under the `docs` node

See https://docs.microsoft.com/en-us/rest/api/searchservice/search-documents#bkmk_examples for example queries.


lawrencegripper:
Looks good - please update help to show keyboard shortcut

PR #156, opened 2019-11-08
DRAFT: Basic fuzzer to talk randomly until crash

lawrencegripper:
Aim is for all PRs to be run under this for a while (maybe 15mins) to see if anything breaks. 

It's non deterministic atm might try and fix that so it walks more logically. 

Need to get it to record nice history/log file for debugging


lawrencegripper:
See #178

PR #160, opened 2019-11-12
Generalise command palette

stuartleeks:
Previous behaviour required the command palette to have a prefix for operations (e.g. "/" for filtering, "search:" for azure search queries). New behaviour splits picking the command from executing the command.
Created a command interface to allow key handlers to also be commands where this makes sense. 
Added command implementations for Copy Content, Open Portal, Edit Item
Opening the command palette now prompts for a command from a list (with the ability to type to filter commands)
Minor tweaks (error checking in action command, simpler default command specification for filter)


internal/pkg/keybindings/globalhandlers.go
```diff
@@ -200,44 +214,112 @@ func (h *ClearPendingDeleteHandler) Fn() func(g *gocui.Gui, v *gocui.View) error
...
+	return "Filter"
+}
+func (h *CommandPanelFilterHandler) IsEnabled() bool {
+	return true
+}
+func (h *CommandPanelFilterHandler) Invoke() error {
+	h.commandPanelWidget.ShowWithText("Filter", "", nil, h.CommandPanelNotification)
+	return nil
+}
+func (h *CommandPanelFilterHandler) CommandPanelNotification(state views.CommandPanelNotification) {
```
lawrencegripper: :) My brain is mapping this to somewhere between `INotifyPropertyChanged` and `ICommand` :+1:

lawrencegripper:
Just merging master into this and then hoping to merged @stuartleeks unless you object?

PR #164, opened 2019-11-18
Add support to container registry for deleting tags, manifests and repositories

stuartleeks:


lawrencegripper:
Gonna have a look over this one once I'm back from holidays, if that's alright? Or @jjcollinge can review and merge while I'm away

jjcollinge:
I'll take a look after lunch

stuartleeks:
@jjcollinge - shout if you have any questions

internal/pkg/expanders/containerRegistry.go
```diff
@@ -440,7 +506,7 @@ func (e *ContainerRegistryExpander) getItemsForURL(ctx context.Context, url stri
 	span, ctx := tracing.StartSpanFromContext(ctx, "getItemsForURL(containerregistry):"+url, tracing.SetTag("url", url))
 	defer span.Finish()
 
-	responseBuf, err := e.doRequest(ctx, url, accessToken)
+	responseBuf, err := e.doRequest(ctx, "GET", url, accessToken)
```
jjcollinge: nit: move these verbs to consts

internal/pkg/views/notifications.go
```diff
@@ -112,11 +112,14 @@ func (w *NotificationWidget) ConfirmDelete() {
...
 
-		swaggerExpander := expanders.GetSwaggerResourceExpander()
-
 		for _, i := range pending {
-			swaggerDeleted, err := swaggerExpander.Delete(context.Background(), i)
-			if err == nil && !swaggerDeleted {
+			var err error
+			fallback := true
+			if i.Expander != nil {
+				deleted, err := i.Expander.Delete(context.Background(), i)
```
jjcollinge: general comment: should we be handling contexts a bit better here? Using timeouts and allowing cancellations - something for another PR.

stuartleeks: Yeah, that's probably worth raising an issue for!

internal/pkg/expanders/containerRegistry.go
stuartleeks: I'm not sure how much it adds moving these to consts - I find them quite easy to parse (and locate) as strings. I also had a quick look at the net/http code and that just uses plain strings :-)

stuartleeks:
@jjcollinge are you ok to approve?

PR #165, opened 2019-11-19
Add Copy item ID command

### LINKED ISSUE #21: Feature: Get resource URL

stuartleeks:
Fixes #21 

* Enables commands without a keybinding
* Adds command to copy current item ID

### DIFF
--- a/internal/pkg/keybindings/listhandlers.go
+++ b/internal/pkg/keybindings/listhandlers.go
@@ -748,7 +748,7 @@ func (h ListCopyItemIDHandler) Fn() func(g *gocui.Gui, v *gocui.View) error {
 }
 
 func (h *ListCopyItemIDHandler) DisplayText() string {
-	return "Copy current item ID"
+	return "Copy current resource ID"
 }
 func (h *ListCopyItemIDHandler) IsEnabled() bool {
 	return h.List.CurrentExpandedItem() != nil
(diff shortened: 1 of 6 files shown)


lawrencegripper:
Ps. Command pallet is very very cool now. Like the idea of adding in commands without keyboard shortcuts.

stuartleeks:
@jjcollinge - @lawrencegripper is right that we will have a clash between our PRs. I've moved the clipboard code to a shared method (i.e. encapsulated the wsl vs non-WSL test etc inside a single function) and reused this in another command. Since your PR is marked as draft, are you ok to merge this PR and then fix up in yours?

jjcollinge:
Yeh - I'll fix up once this is merged

stuartleeks:
@jjcollinge - thank you! Are you ok to look at this and approve if you're happy? Then you can update your PR and get that merged (I'm looking forward to that :-))

internal/pkg/keybindings/listhandlers.go
```diff
@@ -722,3 +722,48 @@ func (h *CommandPanelAzureSearchQueryHandler) CommandPanelNotification(state vie
...
+}
+
+func (h ListCopyItemIDHandler) Fn() func(g *gocui.Gui, v *gocui.View) error {
+	return func(g *gocui.Gui, v *gocui.View) error {
+		return h.Invoke()
+	}
+}
+
+func (h *ListCopyItemIDHandler) DisplayText() string {
+	return "Copy current item ID"
```
jjcollinge: nit: I think "Copy current resource ID" is more consistent with Azure terminology.

stuartleeks:
ah, good spot!

stuartleeks:
Ok, rebased on previous merge and fixed up the other messages to use "resource ID"

stuartleeks:
All done - over to you for #166 :-)

Issue #161, opened 2019-11-13
When updating an item the JSON is no longer formatted before getting to the editor


lawrencegripper:
When refactoring I removed some duplicate code which attempted to format each response from the API into an indented JSON. 

As we now have formatting in the `itemview` which behaves differently based on the content type I moved the code. 

The result is that you can view an item with nicely formatted JSON but when you move to updating it you see a single line of JSON as it was returned by the server. 

![editformatting](https://user-images.githubusercontent.com/1939288/68748345-2ccccd80-05f4-11ea-99d4-234a7f449985.gif)

jjcollinge:
Will do this today when I get some time

jjcollinge:
Looks like this also effects saving the json to clipboard

jjcollinge:
#166 should have closed this

PR #166, opened 2019-11-19
Fix JSON formatting #161

jjcollinge:
Fix JSON formatting #161

Not sure this is the best way to fix, just had a quick go at getting something working. Any suggestions.

Using `style/colors.go` seemed to add a bunch of color characters so I just used the vanilla json package. Could extract this common funciton to `style/format.go` or something.

FYI: This will conflict with #165

### DIFF
--- a/internal/pkg/keybindings/globalhandlers.go
+++ b/internal/pkg/keybindings/globalhandlers.go
@@ -1,6 +1,8 @@
 package keybindings
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


lawrencegripper:
Nice this look good to me. 

nit: Should be use the `ContentType` to only invoke the `json` formatting when the response is JSON. Keeps things cleaner if we want to add other formatters for YAML etc. 

Could move this to here: https://github.com/lawrencegripper/azbrowse/pull/166/files#diff-93ef310a1235e982e0a50912580c4827L507-L514

jjcollinge:
> 
> 
> Nice this look good to me.
> 
> nit: Should be use the `ContentType` to only invoke the `json` formatting when the response is JSON. Keeps things cleaner if we want to add other formatters for YAML etc.
> 
> Could move this to here: https://github.com/lawrencegripper/azbrowse/pull/166/files#diff-93ef310a1235e982e0a50912580c4827L507-L514

Yeh makes sense, will update

stuartleeks:
Just created #173 to fix up the merge conflict between this PR and #165

jjcollinge:
FYI: I’m on holiday at the moment and don’t a have a laptop to hand to address this until next Tuesday.

stuartleeks:
Oh wow - you've been busy on your phone then! Have a good holiday 😀

PR #170, opened 2019-11-19
Only show 'waiting' message after waiting

stuartleeks:
Only show the "do you have another instance of Azbrowse open" message after we've actually been waiting to open it

### DIFF
--- a/internal/pkg/storage/store.go
+++ b/internal/pkg/storage/store.go
@@ -6,18 +6,29 @@ import (
 	"github.com/boltdb/bolt"
 	"log"
 	"os/user"
+	"time"
 )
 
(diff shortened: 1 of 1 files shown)


lawrencegripper:
Future work maybe: Now we have a nicer setup in `main.go` do you think it makes sense to move from `init`->`loadDb` can call it explicitly at startup? 

The use of the `init` method was a nasty hack by me

lawrencegripper:
Gonna push that change here as will make me happy - can roll back if you object

stuartleeks:
I've not really looked at the `init` method beyond the changes I made to delay the message display :-)

PR #178, opened 2019-11-26
Deterministic fuzzer (is that an oxymoron?)

lawrencegripper:
The fuzzer in this PR now walks the tree visiting each node in order once. Super-seeds #156 

The aim is to create something that can be easily used when testing and errors can be easily reproduced. 

Features:

- Linked to `-navigate` feature to allow targeted checking of an area/provider you are working on like `azbrowse -fuzzer 1 -navigate /subscriptions/SOMESUBHERE/resourceGroups/lk-scratch`
- Skip/Limit function allows skipping of areas or limited the number of items expanded in an area. For example you can test 1/2 items in the `Activity Log` but you don't want to walk all of them as it would take an age for a resource group which has heavy traffic. 
- Testing function to allow codified tests to be added and asserted by the fuzzer while it's walking the tree. 

Todo:

- [x] Refactor the code into `internal/pkg/automation` (plan to more `-navigate` functionality here too)
- [x] Improve testing and skipping framework
- [ ] Add `fuzz.log` output file to follow the fuzzers walk 
- [x] Add quick output command with pre-populated `-navigate` command on failure to allow quick repro
- [ ] Hook errors from `expanders` and log/assert about them
- [x] Add skip function for `metrics` as walking all is lengthy


lawrencegripper:
@stuartleeks FYI I've been playing a bit in this space, think this is more useful than the previous version - what do you recon?

internal/pkg/automation/fuzzer.go
```diff
@@ -0,0 +1,168 @@
...
+		return b
+	}
+
+	testFunc := func(currentNode *expanders.TreeNode, nodes []*expanders.TreeNode) {
+		// if r := regexp.MustCompile(".*/providers/Microsoft.ContainerRegistry/registries/.*/"); r.MatchString(currentNode.ID) {
+		// 	st.Expect(&ErrorReporter{}, len(nodes), 5)
+		// }
+	}
+
+	shouldSkip := func(currentNode *expanders.TreeNode, itemId string) (shouldSkip bool, maxToExpand int) {
```
jjcollinge: There's currently no circumstances under which shouldSkip is returned true - is this just for future proofing in case you want to add other conditions?

lawrencegripper: Yeah for some more universal providers like `Activity Log` I only really want to ever walk it once no matter how many resources have it. Added this now with a `defer` statement... or attempting to.

lawrencegripper:
I think this is good to go, I've added in more info the `recovery` func so you get details as to where you are when a crash occurs as found I'd come back to a crash and it wouldn't be obvious how the fuzzer made it occur. 

image

PR #179, opened 2019-11-27
Cleanup terminal on crash

lawrencegripper:
This change will catch a `panic` and ensure `azbrowse` returns the terminal to a working state before it closes.

### DIFF
--- a/cmd/azbrowse/main.go
+++ b/cmd/azbrowse/main.go
@@ -78,8 +78,20 @@ func run(settings *Settings) {
 	// Close the span used to track startup times
 	span.Finish()
 
+	// recover from panic if one occurred leaving terminal usable
+	defer func() {
+		if r := recover(); r != nil {
+			g.Close()
+			fmt.Printf("\n\nA crash occurred: %s \n", r)
+			debug.PrintStack()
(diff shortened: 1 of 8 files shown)


stuartleeks:
Ooh, nice!

cmd/azbrowse/main.go
```diff
@@ -78,8 +78,20 @@ func run(settings *Settings) {
...
+			fmt.Printf("\n\nA crash occurred: %s \n", r)
+			debug.PrintStack()
+			fmt.Println("Please visit https://github.com/lawrencegripper/azbrowse/issues to raise a bug.")
+			os.Exit(1)
+		}
+	}()
+
 	// Start the main loop of gocui to draw the UI
 	if err := g.MainLoop(); err != nil && err != gocui.ErrQuit {
+		g.Close()
```
jjcollinge: Is this needed? The line below will always throw a panic that'll be caught by the deferred recover statement which will in turn call `g.Close()` - is there another reason for it?

lawrencegripper: Good call, removed.

lawrencegripper:
Refactored this a bit as I realized the call is required at the start of any go routine that may panic. I've made little helper function to handle this in every case and sorted the formatting to be more friendly on failure. 

image

Issue #177, opened 2019-11-26
Review use of context for resource deletion


jjcollinge:
We're currently using `context.Background()` on both line [119](https://github.com/lawrencegripper/azbrowse/blob/615826ca130c9babafcc3efcfab3d1fc4834f5a4/internal/pkg/views/notifications.go#L119) and line [124](https://github.com/lawrencegripper/azbrowse/blob/615826ca130c9babafcc3efcfab3d1fc4834f5a4/internal/pkg/views/notifications.go#L124) when deleting resources. These should probably have a timeout on the delete call and maybe even support a higher level user invoked cancellation. I'm not 100% sure on how a cancel would behave but I guess it'd be "best effort" i.e. you may have iterated through 3 resources and deleted them but saved the other 7 in the queue if it was accidental. We'd have to add support for the user issuing a cancellation or gracefully killing the program.

lawrencegripper:
Yeah makes sense to me. Good call.

PR #180, opened 2019-11-28
Added contexts to http requests

jjcollinge:
Attach our existing context to HTTP requests and add a timeout to our delete operation.

I've not changed the context in the expanders just attached it to the requests. We [handle timeouts for expanders](https://github.com/lawrencegripper/azbrowse/blob/615826ca130c9babafcc3efcfab3d1fc4834f5a4/internal/pkg/expanders/expander.go#L100) without context.

This probably warrants a bigger discussion about how we configure contexts appropriately for each HTTP request.

issue: #177


lawrencegripper:
Nice looks like a good change to get in and move us in the right direction.

PR #192, opened 2019-12-05
Add argument to specify tenant

stuartleeks:
Add argument to specify which tenant to load subscriptions for and add/update docs

### DIFF
--- a/pkg/armclient/armclient.go
+++ b/pkg/armclient/armclient.go
@@ -33,12 +33,15 @@ type Client struct {
 
 // LegacyInstance is a singleton ARMClient used while migrating to the
 // injected client
-var LegacyInstance = NewClientFromCLI()
+var LegacyInstance Client
 
 // NewClientFromCLI creates a new client using the auth details on disk used by the azurecli
-func NewClientFromCLI() *Client {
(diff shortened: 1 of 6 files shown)


lawrencegripper:
I WANT THIS NOW. LET ME TEST

pkg/armclient/armclient.go
```diff
@@ -33,12 +33,15 @@ type Client struct {
 
 // LegacyInstance is a singleton ARMClient used while migrating to the
 // injected client
-var LegacyInstance = NewClientFromCLI()
+var LegacyInstance Client
 
 // NewClientFromCLI creates a new client using the auth details on disk used by the azurecli
-func NewClientFromCLI() *Client {
+func NewClientFromCLI(subscription string) *Client {
```
lawrencegripper: Nit: this looks like it takes in a `tenantID` so param should this be `tenantID` or have I misunderstood

stuartleeks: Darn - hangover from initial spike on this!

stuartleeks:
Fixed up the misnamed parameter and rebased on latest master :)

Issue #175, opened 2019-11-21
Add more masking to `--demo` mode


stuartleeks:
I think it would be good if the `--demo` mode obscured more fields:

* anything that has a `Secret` suffix under `<web app>/config/authsettings/` 
* `<web app>/config/connectionstrings/` values
* `<web app>/config/appsettings/` values
* `<web app>/config/publishingcredentials/` values: `publishingPassword` and `scmUri`
* `<web app>/config/functions/token` value
* `<azure search>/listAdminKeys` values: `primaryKey` and `secondaryKey`
* `<azure search>/listQueryKeys` all `key` values

A good portion of these would be covered by masking values with suffices of `key`, `secret` or `password`

Thoughts?

stuartleeks:
This could be a "good first issue" once scoped

bengreenier:
I'm going to try and tackle this. First step (as i understand it) is to add the regex values that match each of the examples above to [demomode.go](https://github.com/lawrencegripper/azbrowse/blob/master/internal/pkg/views/demomode.go#L7) if they are not currently matching one of the existing values.

My plan of attack is to go get real values of the above types, generalize them to example values, add these examples to [demomode_test.go](https://github.com/lawrencegripper/azbrowse/blob/master/internal/pkg/views/demomode_test.go), and then add regex to address any failing tests. Hope that sounds reasonable (let me know if I'm way off).

lawrencegripper:
Yup, sounds good to me. Thanks for taking a look at this it :)

bengreenier:
https://github.com/clarkio/azure-mask/blob/master/src/content-script/mask-process.js#L3 might be relevant/interesting. Including a link mostly for my own thought process, or if others are interested.

bengreenier:
We can close this now, I believe

lawrencegripper:
Indeed - thanks!

PR #194, opened 2019-12-06
Feature: Add More demo masks 🎭

bengreenier:
> Note: This has somewhat significant user-facing changes, and I'm not sure how versioning for releases works around these parts, so I'm just calling that out here. 🤠

This addresses #175 - by adding more regex/logic to mask values when running with `--demo`.  🎭

+ Refactor test data to make assertion diffs easier to read (dev-xp)
+ Move test data after test body (dev-xp)
+ Add tests for new cases
    - webapp/config/authsettings
    - webapp/config/connectionstrings (if ending with key, secret, password suffix)
    - webapp/config/appsettings
    - webapp/config/publishingcredentials
    - webapp/functions/admin/token
    - search/listQueryKeys
    - search/listAdminKeys
+ Add test enforcing json data (only for our test data)
    - this is helpful as there seems to have been [a bug with extra commas](https://github.com/lawrencegripper/azbrowse/blob/master/internal/pkg/views/demomode_test.go#L17) in our test data
    - there has been no change to stripSecretVals, this is purely a test against our own data set for sanity
+ Add logic to support `type` and `name` masking conditions
+ Add regex for new mask cases
+ Add regex for masking of any value with a suffix `[kK]ey`, `[sS]ecret`, or `[pP]assword`

<details>
<summary>
Some screenshots (with my user data now obfuscated!)
</summary>

function_connstr

function_token

function_pubcred

</details>


internal/pkg/views/demomode.go
```diff
@@ -20,13 +45,41 @@ func stripSecretVals(s string) string {
...
+	anyKeyRegex := regexp.MustCompile(`(".*[kK]ey":\s*").+?(?:\\"|[^"])*(")`)
+	s = anyKeyRegex.ReplaceAllString(s, "${1}HIDDEN-KEY${2}")
+
+	anySecretRegex := regexp.MustCompile(`(".*[sS]ecret":\s*").+?(?:\\"|[^"])*(")`)
+	s = anySecretRegex.ReplaceAllString(s, "${1}HIDDEN-SECRET${2}")
+
+	anyPasswordRegex := regexp.MustCompile(`(".*[pP]assword":\s*").+?(?:\\"|[^"])*(")`)
+	s = anyPasswordRegex.ReplaceAllString(s, "${1}HIDDEN-PASSWORD${2}")
+
 	sshRegex := regexp.MustCompile(`ssh-rsa AAAA[0-9A-Za-z+/]+[=]{0,3} ([^@]+@[^@]+)`)
```
lawrencegripper:
```suggestion
	sshRegex := regexp.MustCompile(`ssh-rsa AAAA[0-9A-Za-z+/]+[=]{0,3} ([^@]+@[^@]+)?`)
```

This isn't part of the PR - noticed while testing that an SSH key I have didn't get stripped as it didn't have an email address.

lawrencegripper: Probably better to track in a separate issue and I can add a test too.

lawrencegripper:
This is awesome @bengreenier thanks for all the hard work!

Re: versioning I've been fairly lax, each release is just a `v1.1.{BUILDID}` we could bump the version here.. probably long overdue for `v1.2.x`

https://github.com/lawrencegripper/azbrowse/blob/35560cb54a4bceefabe7abed17a3e4ef4cc800c2/scripts/release.sh#L10

internal/pkg/views/demomode.go
bengreenier: I think i'll just do this here if that's okay (incl test) that way it's in the same version bump. I'll proceed as if you've said "yes Ben that's great please do that" unless I hear otherwise :)

bengreenier:
Updated ssh-key regex as well, based on feedback from @lawrencegripper 

Regex validation can be seen here: https://regex101.com/r/gNqCAN/2/tests

lawrencegripper:
@bengreenier the regex101 test stuff is very cool!

Thanks for the hard work, just merged another PR so let me update and then merge this one.

lawrencegripper:
Looks like Azure devops build was a platform issue or odd edge case - rerunning manually on the PR succeeded. Will track a task to investigate its oddness.

https://dev.azure.com/lawrencegripper/githubbuilds/_build/results?buildId=655

PR #196, opened 2019-12-10
Skip rendering when doing automated navigation

jjcollinge:
Just a PR exploring skipping content rendering when doing automation tasks such as fuzzing or navigateTo. Not sure this makes sense to merge but just wanted to get a feel for how easy this would be to implement. 

Perf testing didn't show any significant improvement in startup time. Just stops the screen flashing through the different resources.

* Work in progress!


lawrencegripper:
Updating from master so fix in #200 is included.

lawrencegripper:
Update branch to include fix #202 and newer CI

PR #199, opened 2019-12-12
Enable h/j/k/l in addition to cursor keys by default

stuartleeks:


lawrencegripper:
Updating from master so fix in #200 is included.

stuartleeks:
Thanks - are you happy to merge this?

lawrencegripper:
So I've been having a play and have seen an odd behavior which I'd like to understand. 

Say I go up and down a list with `J` x2 then use `L` to switch into the response body on the right I'd then expect `K` to move me back to the list but it looks like once I'm in the right panel only the arrow key works for navigating back as the `K` is being captured by the panel (I think this is won to it being an editor). 

Think we probably need to iron this one out before we ship as guessing this will frustrate `vim` users. 

All that said ... The change to allow multiple bindings for a key is ace and I'm all in favour.

stuartleeks:
I initially thought I'd maybe missed a keybinding, but can't work out what one. The only defaults for cursor keys that I didn't overwrite are for navigating up and down the command panel (and since that allows type to filter, `j`/`k` aren't so great to bind there)

I've just spent a bit of time trying to repro that and navigating seems to be working fine - I can move up and down the list with `j`/`k`, move to the right panel with `l` and move back to the list with `h`

If you manage to repro then let me know the steps and I'll take a look

lawrencegripper:
Sooooooooo I'm an idiot and go muddled about my VIM key bindings ... really sorry!

stuartleeks:
Lol, no worries. I looked them up to be sure when implementing!

Are you ok to merge then?

lawrencegripper:
Yeah lets do it. I'll track an issue for the VIM style navigation in the `itemview` window and have a think on how I could tackle it as a separate PR.

PR #204, opened 2020-01-02
Move to full path comparison during sort

### LINKED ISSUE #202: Swagger-gen non-deterministic output (unsure)
I'm seeing a weird behavior and tracking here to come back to it. 

It looks like I can end up with elements of the `swagger-generated` file appearing if different orders on subsequent runs:

image

My first reading of this is that the ordering has returned differently for some reason and so the file has been created, essentially the same, but with a different order of elements. 

Will continue to dig.

lawrencegripper:
Fix #202 

This is a draft showing moving to using the full `templatedUrl` for swagger sort.

### DIFF
--- a/internal/pkg/expanders/swagger-armspecs.generated.go
+++ b/internal/pkg/expanders/swagger-armspecs.generated.go
@@ -1017,6 +1017,10 @@ func (e *SwaggerAPISetARMResources) loadResourceTypes() []swagger.ResourceType {
 			Endpoint: endpoints.MustGetEndpointInfoFromURL("/subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Insights/components", "2015-05-01"),
 			SubResources: []swagger.ResourceType{
 				{
+					Display:  "webtests",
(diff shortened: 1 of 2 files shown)


internal/pkg/expanders/swagger-armspecs.generated.go
```diff
@@ -5951,29 +5951,29 @@ func (e *SwaggerAPISetARMResources) loadResourceTypes() []swagger.ResourceType {
 		},
 		{
 			Display:  "{applyUpdateName}",
-			Endpoint: endpoints.MustGetEndpointInfoFromURL("/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/{providerName}/{resourceType}/{resourceName}/providers/Microsoft.Maintenance/applyUpdates/{applyUpdateName}", "2018-06-01-preview"),
+			Endpoint: endpoints.MustGetEndpointInfoFromURL("/subscriptions/{subscriptionId}/resourcegroups/{resourceGroupName}/providers/{providerName}/{resourceParentType}/{resourceParentName}/{resourceType}/{resourceName}/providers/Microsoft.Maintenance/applyUpdates/{applyUpdateName}", "2018-06-01-preview"),
```
stuartleeks: It looks like the hierarchy generation code is failing on these ones, but that is a pre-existing problem!

lawrencegripper: Good spot, tracked in #205

Issue #15, opened 2018-12-19
Feature: When performing async actions notify on completion


lawrencegripper:
For example when doing a delete it would be nice to get notified when the delete is completed.

lawrencegripper:
Details on how the ARM client will behave: https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-manager-async-operations

Hook for the check of status code: https://github.com/lawrencegripper/azbrowse/blob/e85ea23/pkg/armclient/armclient.go#L131

Example on how to send a statusEvent to show in the UI once something is complete: https://github.com/lawrencegripper/azbrowse/blob/e85ea23/internal/pkg/views/notifications.go#L47

lawrencegripper:
Having a go at this now.

lawrencegripper:
Completed in #206 (outstanding bug #217)

PR #206, opened 2020-01-02
Support Toast notifications for async actions like `Delete` or `Restart`

lawrencegripper:
#15 

Bonus, fix some buggy code in the `eventing` package and update some docs. 

Todo: 
- [ ] ~~Continue refactoring of `eventing.SendStatusEvent`~~ For another PR
- [x] Improve test coverage (integration tests)
- [x] Check behavior on things like `restart` and `shutdown`

### DIFF
--- a/internal/pkg/views/statusbar.go
+++ b/internal/pkg/views/statusbar.go
@@ -123,13 +123,12 @@ func (w *StatusbarWidget) Layout(g *gocui.Gui) error {
 		w.currentMessage.Message = stripSecretVals(w.currentMessage.Message)
 	}
 
-	if w.currentMessage.InProgress {
+	if w.currentMessage.InProgress || w.currentMessage.Failure {
 		fmt.Fprint(v, style.Loading(w.currentMessage.Icon()+"  "+w.currentMessage.Message))
-	} else if w.currentMessage.Failure {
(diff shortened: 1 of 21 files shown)


internal/pkg/views/statusbar.go
```diff
@@ -124,11 +124,11 @@ func (w *StatusbarWidget) Layout(g *gocui.Gui) error {
 	}
 
 	if w.currentMessage.InProgress {
-		fmt.Fprint(v, style.Loading("⏳  "+w.currentMessage.Message))
+		fmt.Fprint(v, style.Loading(w.currentMessage.Icon()+"  "+w.currentMessage.Message))
```
jjcollinge: This condition has been moved into the `Icon()` method so arguably isn't required (atleast for the `InProgress` and `Failure`, might still want then to apply a different style to `Completed`)

lawrencegripper: Good spot - will fix up. Thanks!

lawrencegripper: Done - thanks!

lawrencegripper:
I'm much happier with this now, only remaining work is to remove the panics and improve the test coverage.

lawrencegripper:
I'm pretty happy this is in a fairly good working state. It's a pretty broad PR now so my preference would be to merge and revisit the remaining work in a separate PR

PR #211, opened 2020-01-07
bugfix(itemview): Use keyed literals

### LINKED ISSUE #193: Bug: Devcontainer linter failure - composites
Hello, I'm not actually sure this is isolated to the devcontainer, but i suspect it is based on previous conversations we've had.

```
github.com/alecthomas/chroma.Rule composite literal uses unkeyed fieldscomposites
```

Which appears to be caused by [this structure](https://github.com/lawrencegripper/azbrowse/blob/master/internal/pkg/views/itemview.go#L184) - specifically all 19 similar structures in the same file near that area.

image

I suspect, the `//nolint:govet` comments are supposed to be preventing this type of warning, but aren't working for some reason (I'm not familiar enough with this stuff to diagnose the issue). That said - One [potential solution](https://stackoverflow.com/a/54548495) seems to be to simply disable that type of warning globally, but I'd assume that isn't desirable for the project.

bengreenier:
Fixes devcontainer linter warning about unkeyed composites in a dependency-defined structure. This is unsafe, because the dependency could change field order, possibly while still being syntactically valid, yeilding inconsistent behaviour.

Fixes #193

### DIFF
--- a/azure-pipelines.yaml
+++ b/azure-pipelines.yaml
@@ -2,7 +2,15 @@ pool:
   vmImage: 'Ubuntu 18.04'
 
 steps:
-- script: sudo -E make devcontainer-release
+- script: |
+    # WAT? Yeah if the build doesn't have access to the variable it sets it to a litaral of the variable name with dollar sign and all
+    # this then causes make to fail so we find these and unset them. So GITHUB_TOKEN='$GITHUB_TOKEN' if we don't remove it. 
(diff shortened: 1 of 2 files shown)


bengreenier:
Rebased onto master

bengreenier:
Makefile:80: *** Recursive variable 'GITHUB_TOKEN' references itself (eventually).  Stop.

lawrencegripper:
Let me look at the build failure.

PR #212, opened 2020-01-07
Bugfix(gopls): Prevent warning about outdated gopls

### LINKED ISSUE #209: Devcontainer gopls out of date
The Go extension complains about out of date `gopls` when opening the container.

> Note: If you choose to update, it seems to persist it into the container (until it is rebuilt), so it's not everytime the project is opened.

image

As I understand it, `gopls` is alpha, so maybe this pinned version is intentional and the error should be ignored?

Seems there's 1 of 2 actions to take here:
+ Document that you should close/ignore the popup error
+ Update the Dockerfile to use [`gopls@v0.2.2`](https://github.com/golang/tools/releases/tag/gopls%2Fv0.2.2)

bengreenier:
This prevents warning with a popup in vscode when gopls is not running the latest version. We pin our gopls version for consistency, and don't want to have to deal with this popup each time the container is used.

Fixes #209

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,6 +4,7 @@
         "--fast"
     ],
     "go.useLanguageServer": true,
+    "go.useGoProxyToCheckForToolUpdates": false,
     "go.inferGopath": false,
     "go.delveConfig": {
         "dlvLoadConfig": {


lawrencegripper:
Merged master into this PR to pull in the build fix.

PR #223, opened 2020-02-29
Re-enable "Update current item" command

### LINKED ISSUE #222: Update command to edit resources not working
On latest master, the "Update current item" command is not enabled even on nodes that support editing, and the keyboard shortcut has no effect

stuartleeks:
Leaf node navigation now stores the node on the Content view and the ListUpdate command now checks the Content view node. This allows resource editing to work without reverting to behaviour that allows infinite expansion of a leaf node.

PR also changes `make debug` to use `dlv debug` as this gives a richer debug experience compared to debugging the compiled binary. Happy to pull this into a separate PR if preferred

Fixes #222

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ build: swagger-codegen test checks
 ## debug:
 ##		Starts azbrowse using Delve ready for debugging from VSCode.
 debug:
-	GO111MODULE=on go build ./cmd/azbrowse &&  dlv exec ./azbrowse --headless --listen localhost:2345 --api-version 2
+	GO111MODULE=on dlv debug ./cmd/azbrowse --headless --listen localhost:2345 --api-version 2
 
 ## dcterminal 
 ##		Starts an interactive terminal running inside the devcontainer
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,11 @@ build: swagger-codegen test checks
 debug:
 	GO111MODULE=on dlv debug ./cmd/azbrowse --headless --listen localhost:2345 --api-version 2
 
+## debug-fuzzer:
+##		Starts azbrowse using Delve ready for debugging from VSCode and running the fuzzer.
+debug-fuzzer:
+	GO111MODULE=on dlv debug ./cmd/azbrowse --headless --listen localhost:2345 --api-version 2 -- -fuzzer 5
+
 ## dcterminal 
 ##		Starts an interactive terminal running inside the devcontainer
 dcterminal:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -35,7 +35,7 @@ build: swagger-codegen test checks
 ## debug:
 ##		Starts azbrowse using Delve ready for debugging from VSCode.
 debug:
-	GO111MODULE=on go build ./cmd/azbrowse &&  dlv exec ./azbrowse --headless --listen localhost:2345 --api-version 2
```
lawrencegripper: Nice!

internal/pkg/views/list.go
```diff
@@ -254,31 +254,30 @@ func (w *ListWidget) Navigate(nodes []*expanders.TreeNode, content *expanders.Ex
...
 	if len(nodes) > 0 {
 		w.expandedNodeItem = currentItem
 		w.SetNodes(nodes)
-	}
 
-	w.contentView.SetContent(content.Response, content.ResponseType, title)
-	if currentItem != nil {
-		w.title = w.title + ">" + currentItem.Name
+		if currentItem != nil {
+			w.title = w.title + "fooo>" + currentItem.Name
```
lawrencegripper:
```suggestion
			w.title = w.title + ">" + currentItem.Name
```

lawrencegripper: image

Nit: guess this was for testing

internal/pkg/keybindings/listhandlers.go
```diff
@@ -496,7 +496,7 @@ func (h *ListUpdateHandler) IsEnabled() bool {
 	return true
 }
 func (h *ListUpdateHandler) Invoke() error {
-	item := h.List.CurrentExpandedItem()
+	item := h.Content.GetNode()
 	if !h.IsEnabled() {
 		return nil
```
lawrencegripper:
```suggestion
		eventing.SendStatusEvent(&eventing.StatusEvent{
			InProgress: false,
			Failure:    true,
			Message:    "Updating not supported on this item",
			Timeout:    time.Duration(time.Second * 2),
		})
		return nil
```

internal/pkg/views/list.go
lawrencegripper: Fixed

internal/pkg/keybindings/listhandlers.go
lawrencegripper: Added

lawrencegripper:
Found the issue with the fuzzer, all nodeIds were coming through as root so causing the state to get confused. All fixed up now, will merge once build is passed.

stuartleeks:
Thanks for the fixes :-)

PR #224, opened 2020-02-29
Update swagger processing

stuartleeks:
It turns out that my previous assumptions about the swagger-specs repo were invalid!

Previously the `swagger-update` crawled the folders to find the version folder (e.g. `2019-10-01`) with the latest name. However, the folders don't always contain the full set of files and can just be the files that changed since the previous version. The definition of files that are in an api version are defined in yaml embedded in a `readme.md` 

This PR converts the `swagger-update` process from a bash script to python script and updates the logic based on the above discoveries.


stuartleeks:
UPDATE:

* refactored the code to try to make the swagger-update code more readable
* rebased to split into commits to try to make it easier to view the changes
  * `Convert swagger update to python` - this is the refactoring of the bash script to python
  * `swagger-update` - results of running `make swagger-update` to update the `swagger-specs` folder
  * `swagger-codegen` - results of running `make swagger-codegen`
* rebased on master

lawrencegripper:
Awesome, is now a good time to take a look through the code?

stuartleeks:
Yup - that would be great :-)

scripts/swagger_update/test_swagger_update.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+from swagger_update import *
```
lawrencegripper, line 2:
```suggestion
from api_set import *
```

scripts/swagger_update/test_swagger_update.py
```diff
@@ -0,0 +1,53 @@
...
+
+def test_get_api_version_from_readme_with_invalid_file_path_returns_none():
+    result = get_api_version_from_readme(
+        "./scripts/swagger_update/test_data/does_not_exist.md"
+    )
+    assert result == None, "Should return None for file that doesn't exist"
+
+
+def test_get_api_version_from_readme_with_simple_tags_returns_correct_fileset():
+    api_version = get_api_version_from_readme(
```
lawrencegripper, line 13: Looks like this is missing a param:

```
TypeError: get_api_version_from_readme() missing 2 required positional arguments: 'readme_path' and 'version_overrides'
```

stuartleeks:
Apologies - not having the swagger update tests in `make test` meant that I wasn't pulled up when I forgot to update the tests after refactoring :disappointed: 

Fixed now!

PR #227, opened 2020-03-10
Feature: azfs

lawrencegripper:
Can I use the azbrowse codebase to make a fuse based filesystem which lets you browse and edit azure resources?

[Edit] Yeah you can
azfs

Todo hack:
- [x] root node loading
- [x] walk down into RG and beyond
- [x] delete support
- [x] edit support

Todo stabilize:
- [x] Build out some basic integration tests
- [x] Allow delete without corrupted file system notice
- [x] Allow direct navigation like `ls /sub/rg/resource/subresource` current have to do `ls /sub/`, `ls /sub/rg` etc all the way down the tree
- [x] Throttle API Call rate
- [x] ~~Prevent things like vscode walking the entire tree (every node) by lazy loading or something similar~~ This looks like its a users/app concern. If they want to walk the tree the filesystem should let them. 
- [x] Allow editing files with out `io error` on save

Chore:
- [x] docs
- [x] Fix windows build (aka remove azfs when building for windows)


jjcollinge:
interesting...

lawrencegripper:
azfs

Very early days but it does work - the question is ... is this a useful way to browse and interact with arm?

cc @stuartleeks

bengreenier:
Seems cool to me - more useful than other mechanisms? Not sure. Maybe like pipe in some app settings...

jjcollinge:
Would be cool to be able to use things like grep and globs to find files or files containing strings... i.e rm *-test. Could even pave the way for cron to run automation scripts to clean up resources etc.

lawrencegripper:
So two fun things: 
1.  Because of @stuartleeks 's awesome stuff with K8s api you can browse into your AKS clusters k8s API as a filesystem 
2.  you can open the thing in vscode 

Caveats: I think it's making way to many calls to the ARM APIs as something in vscode seems to try and walk the whole tree. It's still very unstable.

azfs-k8s-vscode

lawrencegripper:
Got delete working (not sure why I get lots of errors out of the `rm` command need to look into that). 

So now @jjcollinge `rm -r test-*` works 🎉

azfs-delete-stuff

lawrencegripper:
Got editing working (with lots of bugs atm). Thinking I'm calling the experiment a success and maybe now moving to try and work through all the odd kinks of the current code.

lawrencegripper:
So things are going well, I've got a decent set of tests building up against a mocked ARM API and with these in place have been able to fix up lots of bugs. 

As of now:
-  navigation should be simpler and you don't have to walk `/thing1` then `/thing2` you can do `/thing1/thing2`.
- delete should work correctly with `rm -r` without returning errors. Todo: work on test for this behavior as `RemoveAll` in Go and `rm -r` seem to do slightly different things

Hopefully get some time over the weekend to work on the editing stuff.

lawrencegripper:
This seems related in regard to vscode walking the whole directory tree: https://github.com/microsoft/vscode/issues/3703

I'm not sure of an easy way to solve this at the moment, couple of options I'm considering 
1. Detect a pattern that looks like a walk and return empty dirs which are then invalidated. This would prevent calls while allowing the directory to be expanded fully in future
2. Throttle the call rate on the `armclient` only problem here is this wouldn't prevent the directory walk just delay it 
3. Limit the tree depth of the structure making `subResources` files instead of directories. This limits the walk but also limits the usefulness of the filesystem

lawrencegripper:
I've added some basic local rate limiting to make things more sane.

Having looked at the rate limits for ARM API all up I don't see using these as a good guide as by the looks of things they're enforced hourly (need to test) https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling. If that's true we could burn through a users request limit and leave them without access to stuff which would likely annoy them. 

Bit more digging is required here.

lawrencegripper:
Moved this out of draft as now have a nice set of mocked out tests and integration tests and have proved out the use case I wanted it working fairly nicely. 

@stuartleeks / @jjcollinge would be really keen to get a review on this if you have some time. Should be ready to go if you checkout the branch and do `make install && azbrowse azfs -mount /mnt/azfs -accept-risk -sub subherenonproduction` or add `-edit` once you've read the risks in the docs. 

@martinpeck would be interested to know if you can get this working on Mac with Fuse for OSX, non-urgent tho. 

Anyone else willing to have a play is welcome too, please do read the risks before mounting a subscription you have access too.

lawrencegripper:
Should have seen this coming... hmm 
image

PR #243, opened 2020-05-30
Convert command handling to cobra

stuartleeks:
Exploration of https://github.com/spf13/cobra for handling commandline args
* added shell completion
* added a `--tenant-id-from-sub` command to allow identifying the tenant to load via an attached subscription

NOTE: the completion for `--tenant-id-from-sub` doesn't seem to handle spaces in the subscription name very well :-(


stuartleeks:
@jjcollinge IIRC you did a bunch of work on the command handling - would love your feedback on this :-)

lawrencegripper:
Had a quick look through this and really like the idea of auto-completion and the new `tenant-id-form-sub`. 

Would like to have more of a play around with it but in principle I'm really up for this change. I think at the time @jjcollinge refactored the params we talked about whether it was worth using Cobra for them and came to the conclusion that it wasn't worth the added complexity - we didn't think about auto-completion which changes this for quite a bit. 

Do you think it would make sense to move from `tenant-id-for-sub` to just `subscription` then use the existing code combined with the `navigate-to` to drop the user straight into the subscription?

`azbrowse --subscription LGSub...<TAB>` gets auto-completed and then drops you into the subscription when it starts. No worries if this makes things more complex for the logic.

jjcollinge:
> 
> 
> @jjcollinge IIRC you did a bunch of work on the command handling - would love your feedback on this :-)

Sure, I'll try find some time next week to take a look

stuartleeks:
> Do you think it would make sense to move from tenant-id-for-sub to just subscription then use the existing code combined with the navigate-to to drop the user straight into the subscription?
I quite like the idea of the `--subscription` switch - pretty sure I'd make use of that
It's not quite the same as `--tenant-id-from-sub` which loads up all the subscriptions for that tenant, but if it were implemented as a combination of `--tenant-id` and `-navigate <subid>` then a simple back navigation would let you switch subscriptions

stuartleeks:
Updated to `--subscription` instead of `--tenant-id-from-sub` as per previous comments

docs/command-line.md
```diff
@@ -14,6 +14,8 @@ Passing the `--tenant-id` argument allows you to control the Azure Active Direct
 
 Running `az account list --query "[].{name:name, tenantId:tenantId}" -o table` will give you a list of subscriptions and their associated tenant. Then you can pass the tenant to azbrowse, e.g. `azbrowse --tenant-id 00000000-0000-0000-0000-000000000000`
 
+Alternatively you can use the `--tenant-id-from-sub` argument to identify the tenant via a subscription that is linked to it. This can be easier as subscription names tend to be more memorable than tenant GUIDs! For example if subscription `mysub` is linked to the tenant you want to load then you can run `azbrowse --tenant-id-from-sub mysub`.
```
lawrencegripper, line 17: nit: Sorry due to the change I suggested this line needs a tweak like:

```suggestion
Alternatively you can use the `--subscription` argument to launch straight into a Subscription no matter which tentant it it under. With command completion enabled `source <(azbrowse completion bash)` you can use tap to complete partial subscription names. 
```

lawrencegripper:
@stuartleeks this is still called a Spike but if your happy to release them I'm happy to release.

stuartleeks:
Oooh - completion for `--navigate` would be awesome. Might be tricky (at least to do it well!)

And I've renamed the PR :-D

lawrencegripper:
Awesome, merge into master when you ready. I already told twitter - bit overexcited.

stuartleeks:
oops - looks like I messed up the merge (should have done it locally rather than in the browser!)

lawrencegripper:
np, got the fix to push from my end - just go sum getting grumpy.

stuartleeks:
push away!

PR #244, opened 2020-06-08
Support non-default tenant in databricks auth

stuartleeks:
Update the databricks authentication to support authentication against Azure Databricks Workspaces that are not in the user's default tenant (as configured in `az` CLI)


.vscode/settings.json
```diff
@@ -9,7 +9,7 @@
         "dlvLoadConfig": {
             "followPointers": true,
             "maxVariableRecurse": 1,
-            "maxStringLen": 255,
+            "maxStringLen": 8192,
```
lawrencegripper, line 12: Oh such a good call - I should have done this long ago and have no idea why delve defaults to 255 then `...`

PR #249, opened 2020-06-22
Improve startup and autocompletion performance

### LINKED ISSUE #248: Speed up launch time
For the autocompletion feature it'd be great to speed up the time taken to launch the CLI. 

- Currently an update is checked for each time. This can be moved to only check once daily.
- Review use of boltdb and when it is created, can this be improved. 
- Look at critical path to usable browser and what can be done async to speed up loading.

lawrencegripper:
Fixes #248 

- Add environment vars to disable the check or force it
- Tweak devcontainer to fix locale error seen in bash
- Cache the account list from `az` command to improve autocompletion perf

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -60,7 +60,7 @@ ENV GO111MODULE=on
 
 # Configure apt, install packages and tools
 RUN apt-get update \
-    && apt-get -y install --no-install-recommends apt-utils dialog fuse nano xterm \
+    && apt-get -y install --no-install-recommends apt-utils dialog fuse nano xterm locales \
     #
     # Verify git, process tools, lsb-release (common in install instructions for CLIs) installed
     && apt-get -y install git iproute2 procps lsb-release \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -85,6 +85,11 @@ RUN apt-get update \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/*
 
+# Setup locale as required by snapd: https://stackoverflow.com/questions/28405902/how-to-set-the-locale-inside-a-debian-ubuntu-docker-container
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
+
 ENV GIT_PROMPT_START='\033[1;36mazb-devcon>\033[0m\033[0;33m\w\a\033[0m'
 
 # Save command line history 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Currently I'm using it every day **but it is experimental so use with caution on
 
 # Cool what else can it do?
 
-Lots [check out the guided tour here](docs/getting-started.md). For advanced [config review the settings page here](docs/config.mg).
+Lots [check out the guided tour here](docs/getting-started.md). For advanced [config review the settings page here](docs/config.md).
 
 - Edit/Update resource
 - Multi-resource delete
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Currently I'm using it every day **but it is experimental so use with caution on
 
 # Cool what else can it do?
 
-Lots [check out the guided tour here](docs/getting-started.md).
+Lots [check out the guided tour here](docs/getting-started.md). For advanced [config review the settings page here](docs/config.mg).
 
 - Edit/Update resource
 - Multi-resource delete
--- a/cmd/azbrowse/cmd.go
+++ b/cmd/azbrowse/cmd.go
@@ -10,12 +10,16 @@ import (
 	"runtime"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/lawrencegripper/azbrowse/internal/pkg/config"
 	"github.com/lawrencegripper/azbrowse/internal/pkg/filesystem"
+	"github.com/lawrencegripper/azbrowse/internal/pkg/storage"
 	"github.com/lawrencegripper/azbrowse/internal/pkg/tracing"
 	"github.com/spf13/cobra"
 )
+
+const keyForAccountCache = "accountCache"
 
 func handleCommandAndArgs() {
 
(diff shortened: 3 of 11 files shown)


lawrencegripper:
Oooh cool. You can test the custom completion provider by using:

`time azbrowse __complete --subscription LG` [Guide here](https://github.com/spf13/cobra/blob/master/bash_completions.md#debugging)

Looking at timings the `selfUpdate` code isn't getting fired in this flow, which is nice, but does mean the optimisation so far hasn't improved the speed. 

Over 3 runs we can see a rough time pattern for the check. Low `1` sec, High `1.4` sec

```
11:04 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m1.086s
user	0m1.021s
sys	0m0.095s
azb-devcon>/workspaces/azbrowse [lg/startupperf|✚ 1] 
11:04 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m1.400s
user	0m1.329s
sys	0m0.105s
azb-devcon>/workspaces/azbrowse [lg/startupperf|✚ 1] 
11:04 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m1.228s
user	0m1.141s
sys	0m0.117s
```

I'm going to look at caching the results from the call which populates this autocompletion. Thinking even caching for a few mins would greatly improve things as the completion is likely to be called more than once. 

```
out, err := exec.Command("az", "account", "list", "--query", "[].[name, id] | [] | sort(@)", "--output", "tsv").Output()
```
Looking at timing for this call individually it suggest we can save a lot of time here. 

```
azb-devcon>/workspaces/azbrowse [lg/startupperf|✚ 1] 
11:04 # time $(az account list)
A few accounts are skipped as they don't have 'Enabled' state. Use '--all' to display them.
bash: [: too many arguments

real	0m1.304s
user	0m1.206s
sys	0m0.099s
```

lawrencegripper:
Now with the results cached the response time on the first completion call is slow but all others are nice and quick! @stuartleeks the question is now long to set the cache for? At the moment it's 6 hours. One option I thought would be to cache this value for a long time but refresh it in the background each time azbrowse is launched. That way autocompletion is quick and the values don't get stale often. What do you think?

```
11:46 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m1.243s
user	0m1.111s
sys	0m0.146s
azb-devcon>/workspaces/azbrowse [lg/startupperf|✚ 3] 
11:46 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m0.057s
user	0m0.055s
sys	0m0.022s
azb-devcon>/workspaces/azbrowse [lg/startupperf|✚ 3] 
11:46 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m0.054s
user	0m0.075s
sys	0m0.005s
azb-devcon>/workspaces/azbrowse [lg/startupperf|✚ 3] 
11:46 # time $(azbrowse __complete --subscription LG > /dev/null)
Completion ended with directive: ShellCompDirectiveNoFileComp

real	0m0.055s
user	0m0.078s
sys	0m0.007s
```

lawrencegripper:
beforeaftercompletion

stuartleeks:
Nice!

README.md
```diff
@@ -17,7 +17,7 @@ Currently I'm using it every day **but it is experimental so use with caution on
 
 # Cool what else can it do?
 
-Lots [check out the guided tour here](docs/getting-started.md).
+Lots [check out the guided tour here](docs/getting-started.md). For advanced [config review the settings page here](docs/config.mg).
```
stuartleeks, line 20: Typo: 
```suggestion
Lots [check out the guided tour here](docs/getting-started.md). For advanced [config review the settings page here](docs/config.md).
```

.devcontainer/Dockerfile
```diff
@@ -60,7 +60,7 @@ ENV GO111MODULE=on
 
 # Configure apt, install packages and tools
 RUN apt-get update \
-    && apt-get -y install --no-install-recommends apt-utils dialog fuse nano xterm \
+    && apt-get -y install --no-install-recommends apt-utils dialog fuse nano xterm locales \
```
stuartleeks, line 63: Ah, I hit this but didn't get chance to investigate - cool!

cmd/azbrowse/cmd.go
```diff
@@ -98,26 +102,52 @@ func createRootCmd() *cobra.Command {
...
+	}
+}
+
+// Provide support for autocompleting subscriptions
+func subscriptionAutocompletion(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	var accountList []byte
 
+	validCache, value, err := storage.GetCacheWithTTL(keyForAccountCache, time.Hour*6)
+	if !validCache || err != nil {
 		out, err := exec.Command("az", "account", "list", "--query", "[].[name, id] | [] | sort(@)", "--output", "tsv").Output()
```
stuartleeks, line 130: Does it make sense to reuse the `updateAccountCache` function here?

docs/config.md
```diff
@@ -4,6 +4,15 @@ By default azbrowse looks for configuration in `/root/.azbrowse-settings.json` o
 
 Below is a table containing some of the default key bindings. If you'd like to customise the key bindings to be more suitable for your setup, please refer to the section on [custom key bindings](#custom-key-bindings).
 
+## Updates
+
+Users can set Environment Vars to control how AzBrowse checks for updates. 
+
+| Environment Variable | Does |
+| ---------------------| ---------- |
+| `AZBROWSE_FORCE_UPDATE=anyvaluehere` | Force Azbrowse to check for updates on launch. Normaly would only check every 6 hours |
```
stuartleeks, line 13: Typo: 
```suggestion
| `AZBROWSE_FORCE_UPDATE=anyvaluehere` | Force Azbrowse to check for updates on launch. Normally would only check every 6 hours |
```

internal/pkg/storage/store.go
```diff
@@ -82,3 +91,52 @@ func GetCache(key string) (string, error) {
...
+
+	return true, cacheItem, nil
+}
+
+// PutCacheItemForTTL puts an item in the cache bucket
+func PutCacheForTTL(key, value string) error {
+	// Save the Item
+	err := PutCache(key, value)
+	// Track when it was saved
+	errUpdate := PutCache(key+ttlLastUpdatedKey, epocToString())
```
stuartleeks, line 133: Does anything ever clear out values with expired TTLs?

cmd/azbrowse/cmd.go
lawrencegripper: Yes, good catch - I'll fix that up.

internal/pkg/storage/store.go
lawrencegripper: Nope, I initially looked at writing something to do some clearing but felt unnecessary with the current use case as they're the same keys overwritten ever time the app runs. 

The fact that this comes up tho and the naming of the methods means I should put a warning about this in the code as future users might assume that it does. Sound like a good middle ground?

cmd/azbrowse/cmd.go
lawrencegripper: Done, just giving it a quick test

internal/pkg/storage/store.go
lawrencegripper: Note added to function description

PR #250, opened 2020-06-24
Autocompleting navigate command

lawrencegripper:
autocompleteNavigate


lawrencegripper:
Fixes #251 too while I'm here.

PR #255, opened 2020-06-25
Moving to diskkv to support multiple instances

### LINKED ISSUE #253: Remove limit of only one running instance
Currently BoltDB, used for caching, limits us to only having one instance open at a time. 

This would look at removing this limitation by using `sqlite` or just flat files to provide the same storage. 

Investigation needed.

lawrencegripper:
Closes #253


lawrencegripper:
// Todo tidy error handling. Rationalize storage location for config and db

PR #277, opened 2020-09-01
Add option to resume from last session

### LINKED ISSUE #252: Add a `--resume` command to restart at last exit point
@stuartleeks came up with this idea and I like it. 

1. Navigate into the subscription/resources. 
2. Exit to do something else. 
3. Use `azbrowse --resume` to jump back to previously open point. (using existing --navigate functionality)

lawrencegripper:
Closes #252 allowing `azbrowse --resume` or `-r` to start where you left off in the last session.

### DIFF
--- a/cmd/azbrowse/cmd.go
+++ b/cmd/azbrowse/cmd.go
@@ -100,12 +100,12 @@ func createRootCmd() *cobra.Command {
 					fmt.Println("Cannot resume: " + err.Error())
 					os.Exit(1)
 				}
-				tenantID, err := storage.GetCache(resumeTenantIDKey)
+				currentTenantID, err := storage.GetCache(resumeTenantIDKey)
 				if err != nil {
 					fmt.Println("Cannot resume: " + err.Error())
 					os.Exit(1)
 				}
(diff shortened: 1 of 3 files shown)


internal/pkg/views/list.go
```diff
@@ -298,19 +298,17 @@ func (w *ListWidget) Navigate(nodes []*expanders.TreeNode, content *expanders.Ex
...
 	w.contentView.SetContent(currentItem, content.Response, content.ResponseType, title)
 	if len(nodes) > 0 {
-		if w.currentPage != nil {
-			w.currentPage.ExpandedNodeItem = currentItem
-		}
 		w.SetNodes(nodes)
 
 		if currentItem != nil {
 			w.currentPage.Title = titlePrefix + currentItem.Name
+			w.currentPage.ExpandedNodeItem = currentItem
```
lawrencegripper, line 305: Note this change may be better suited as part of #276

lawrencegripper: Sorted with rebase now that other PR is merged.

PR #281, opened 2020-10-25
Initial Blob storage implementation

stuartleeks:
Browse blobs and metadata
Delete blobs

### DIFF
--- a/internal/pkg/expanders/storageBlob.go
+++ b/internal/pkg/expanders/storageBlob.go
@@ -1,0 +1,682 @@
+package expanders
+
+import (
+	"bytes"
+	"context"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/xml"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/lawrencegripper/azbrowse/internal/pkg/tracing"
+	"github.com/lawrencegripper/azbrowse/pkg/armclient"
+)
+
+// NewStorageBlobExpander creates a new instance of StorageBlobExpander
+func NewStorageBlobExpander(armclient *armclient.Client) *StorageBlobExpander {
+	return &StorageBlobExpander{
+		client:    &http.Client{},
+		armClient: armclient,
+	}
+}
+
+// Check interface
+var _ Expander = &StorageBlobExpander{}
+
+// StorageListKeyResponse is used to unmarshal a call to listKeys on a storage account
+type StorageListKeyResponse struct {
+	Keys []struct {
+		KeyName     string `json:"keyName"`
+		Value       string `json:"value"`
+		Permissions string `json:"permissions"`
+	} `json:"keys"`
+}
+
+// StorageAccountResponse is a partial representation of the storage account response
+type StorageAccountResponse struct {
+	Properties struct {
+		PrimaryEndpoints struct {
+			Blob string `json:"blob"`
+			Dfs  string `json:"dfs"`
+		} `json:"primaryEndpoints"`
+	} `json:"properties"`
+}
+
+// ContainerListResponse is a partial representation of the List container response
+type ContainerListResponse struct {
+	XMLName    xml.Name `xml:"EnumerationResults"`
+	Blobs      []Blob   `xml:"Blobs>Blob"`
+	NextMarker string   `xml:"NextMarker"`
+}
+
+type Blob struct {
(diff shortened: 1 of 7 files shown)


internal/pkg/expanders/storageBlob.go
```diff
@@ -0,0 +1,682 @@
...
+
+		nodes = append(nodes, node)
+	}
+	if response.NextMarker != "" {
+		node := TreeNode{
+			Parentid:  currentItem.ID,
+			Namespace: "storageBlob",
+			ID:        currentItem.ID + "/" + "...more",
+			Name:      "more...",
+			Display:   "more...",
```
lawrencegripper, line 322: oh this is a very cool solution

internal/pkg/expanders/storageBlob.go
```diff
@@ -0,0 +1,682 @@
...
+	return containerName
+}
+
+func (e *StorageBlobExpander) getAccountKey(ctx context.Context, containerID string) (string, error) {
+
+	i := strings.Index(containerID, "/blobServices")
+	rootURL := containerID[0:i]
+	listKeysURL := rootURL + "/listKeys?api-version=2019-06-01"
+
+	data, err := e.armClient.DoRequest(ctx, "POST", listKeysURL)
```
lawrencegripper, line 442: Would it make much difference to cache this result? We have the storage api with a ttl cache (or rough approximation of one) which could be used. Could key off the storage endpoint or accountid. 

Haven't dug into how regularly this call is made, if it's once up front then not worth it.

internal/pkg/expanders/storageBlob.go
```diff
@@ -0,0 +1,682 @@
...
+
+	return response.Keys[0].Value, nil
+}
+func (e *StorageBlobExpander) getStorageBlobEndpoint(ctx context.Context, containerID string) (string, error) {
+
+	i := strings.Index(containerID, "/blobServices")
+	rootURL := containerID[0:i]
+	storageAccountURL := rootURL + "?api-version=2019-06-01"
+
+	data, err := e.armClient.DoRequest(ctx, "GET", storageAccountURL)
```
lawrencegripper, line 465: Similar note on caching, not sure how relevant it is, also might be something to track in an issue and tackle as a separate PR.

internal/pkg/expanders/storageBlob.go
```diff
@@ -0,0 +1,682 @@
...
+		"", // Empty date because x-ms-date is expected (as per web page above)
+		headers.Get(headerIfModifiedSince),
+		headers.Get(headerIfMatch),
+		headers.Get(headerIfNoneMatch),
+		headers.Get(headerIfUnmodifiedSince),
+		headers.Get(headerRange),
+		buildCanonicalizedHeader(headers),
+		canonicalizedResource,
+	}, "\n")
+	return stringToSign, nil
```
lawrencegripper, line 612: I see you've been having all the fun!

internal/pkg/keybindings/listhandlers.go
```diff
@@ -536,8 +537,12 @@ func (h *ListUpdateHandler) Invoke() error {
 		formattedContent = formattedBuf.String()
 	case expanders.ResponseYAML:
 		fileExtension = ".yaml"
-
 		formattedContent = content // TODO: add YAML formatter
+
+	case expanders.ResponseXML:
+		fileExtension = ".xml"
+		formattedContent = xmlfmt.FormatXML(content, "", "  ")
+
```
lawrencegripper, line 545: Oh nice adding it to the formatter.

lawrencegripper (approved):
LGTM

internal/pkg/expanders/storageBlob.go
lawrencegripper: nit: My expectation was to see the new items added to the existing list rather than stack a new page. Practically speaking tho I think this would be hard with the way expanders are built at the moment and this is a nice way to get the functionality in.
image

internal/pkg/expanders/storageBlob.go
lawrencegripper: Playing around with the code I didn't see any lag waiting on these calls so think this is me over-optimising.

internal/pkg/expanders/storageBlob.go
stuartleeks: There's a precedent for the current approach in the container registry expander. I agree that it would be nice to extend the existing list and added #282

internal/pkg/expanders/storageBlob.go
stuartleeks: Let's see how it goes with a bit of usage and address if needed?

Issue #278, opened 2020-09-01
Status bar lags when changes occur in the UI


lawrencegripper:
When doing multiple navigations in quick succession the status bar appears to lag behind, need to investigate more.

lawrencegripper:
Taking a look at this now, think the existing code needs a rethink and some test cases

lawrencegripper:
Fixed up and bonus better CPU usage when idle.

PR #287, opened 2020-12-22
Fix status bar lag, optimise UI redraw

lawrencegripper:
This PR ensures that the `statusbar` messages appear in a timely fashion and the correct messages show with loading indicators during loading.

To do this it:
1. Reworks the logic which decides which message to show
2. Moves code which blocked the gocui drawing go routine into it's own goroutine. Note this means we now have locking implemented in `list.go` to ensure only one navigation takes place at a time.
3. Uses basic change tracking in update loops to ensure we only redraw if we have too (previously would redraw every second regardless)

Added benefit of this work is reduced CPU usage when idle.

### DIFF
--- a/cmd/azbrowse/main.go
+++ b/cmd/azbrowse/main.go
@@ -178,25 +178,22 @@ func startPopulatingList(ctx context.Context, g *gocui.Gui, list *views.ListWidg
 
 		done()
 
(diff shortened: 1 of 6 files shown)


lawrencegripper:
closes #278

PR #288, opened 2020-12-26
Update swagger specs for azure apis

lawrencegripper:
This uses the work of @stuartleeks to pull down the latest set of of API specs and use them in AzBrowse.

I used overrides for `storage` and `azsadmin` both of which had files missing from their `input` sections. I also added a fix for a `common.json` file which had capitalization so was `Common.json` (yey!).

As a fix I also removed the `get-pip.py` as newer docker image installs pip via apt. (I hope)


lawrencegripper:
Going to wait on the build and give this a try tomorrow morning with `make fuzz` to see if it's thrown up any oddities during the generation.

lawrencegripper:
I've removed the `storage` override in favour of a new `include_additions` property. This allows you to add an arbitrary filepath to be included in the `swagger-specs` folder, in the case of storage this is used to ensure `privatelinks.json` is copied even through it is missing from the `input-files` in the `readme.md`

PR #290, opened 2021-01-05
Move action discovery to be part of expanders

### LINKED ISSUE #289: Move action discovery to be part of expanders
Moving action discovery to be part of expanders would allow an expander to add custom actions, for example unlocking a storage blob.

stuartleeks:
Closes #289

* Refactoring and updates to retrieve actions from expanders
* Previous actions now returned from `DefaultExpander`
* Storage blob now has Acquire/Break lease actions
* Storage blob metadata approach updated to allow refreshing


lawrencegripper (approved):
LGTM! Awesome work

lawrencegripper (approved):

PR #291, opened 2021-01-13
Add "Get Terraform" action

stuartleeks:
Adds "Get Terraform" action to resource nodes to get the Terraform snippet for creating the current resources
Adds "Get Terraform (recursive)" action 

WIP:
 - need to expand the list of supported resource types --> think there's a good starting point now and more can be added in future PRs
 - ~~currently using a depth limit of 5 when recursing the tree. Need to decide whether this is sufficient or whether there's a better way to handle this~~


stuartleeks:
@lawrencegripper I've pulled in the tfprovider code from https://github.com/lawrencegripper/tfoperatorbridge/ and made some changes to integrate into azbrowse. I've also copied and updated the tests, but there's currently a failure as one of the tests requires Azure credentials to run. What are your thoughts on this? Are you ok to add that to the azbrowse pipeline credentials, or shall I remove that test?

internal/pkg/expanders/containerRegistry.go
```diff
@@ -678,7 +678,3 @@ func (e *ContainerRegistryExpander) getRegistryToken(ctx context.Context, loginS
 
 	return accessToken, nil
 }
-
-func (e *ContainerRegistryExpander) testCases() (bool, *[]expanderTestCase) {
```
lawrencegripper, line 682: Nice tidying!

stuartleeks: Thanks - I felt a bit bad bundling it in this PR, but.... 😉

lawrencegripper (approved):
LGTM

lawrencegripper:
This is very cool!!!!! Awesome work sir

stuartleeks:
Thanks - still more that I'd like to do but happy with the progress so far 😄

PR #294, opened 2021-01-19
Format on SetContent

### LINKED ISSUE #292: Scrolling pgDown not working correctly on RHS panel
When viewing Activity Log output within a Resource Group, the content on the right hand side does not scroll correctly.

pgDown key not working.  If you use the down arrow key to scroll past the first few lines.  Press pgDown again and it will return you to the beginning of the document?!

image

Now use the down arrow key to scroll further into the document and use the pgUp key, which will work as expected.

Tested with multiple resource groups, each exhibiting the same pgDown key problem when viewing the Activity Log content

stuartleeks:
Should improve performance
closes #292 PageDn failing on single line content

There's still room for improvement as the PageDown handling computes the number of lines based on the count of `\n` characters in the content, and doesn't account for line wrapping

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ build: swagger-codegen checks test
 ## debug:
 ##		Starts azbrowse using Delve ready for debugging from VSCode.
 debug:
-	dlv debug ./cmd/azbrowse --headless --listen localhost:2345 --api-version 2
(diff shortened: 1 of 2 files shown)


lawrencegripper:
> There's still room for improvement as the PageDown handling computes the number of lines based on the count of \n characters in the content, and doesn't account for line wrapping

You sir are a genius this has been bugging me for AGES!

lawrencegripper (approved):
LGTM - Much nicer!

PR #295, opened 2021-01-25
Make provider lookup case insensitive and update CosmosDB api version

stuartleeks:
- update swagger config to pick up latest Cosmos DB version
 - update api version lookup to be case-insensitive on resource type (fixes issue preventing CosmosDB node expanding)
 - update swagger-update to use `https://` repo address rather than `git@...` repo address to work with anonymous access


scripts/swagger_update/app.py
```diff
@@ -29,6 +29,9 @@
         # storage package-2019-06 references privatelinks.json which is not listed in the input files
         "storage": { 
             "package-2019-06": ["./Microsoft.Storage/stable/2019-06-01/privatelinks.json"] 
+        },
+        "cosmos-db" : {
+            "package-2020-04" : ["./Microsoft.DocumentDB/stable/2019-08-01/cosmos-db.json"]
```
lawrencegripper, line 34: wooo I did something useful!

lawrencegripper (approved):
LGTM

scripts/swagger_update/app.py
stuartleeks: YES - very much YES!

stuartleeks: Seeing that made me happy as I was working on this 😄

PR #303, opened 2021-02-23
Add "Get Document" action to Cosmos DB data plane

stuartleeks:
Adds a "Get Document" action to Cosmos DB collection nodes
Changes the status colour from red to yellow for in-progress notifications


lawrencegripper (approved):
LGTM

PR #310, opened 2021-03-12
Move to supported cgoui fork

lawrencegripper:
Get
- concurrent ui updates
- tcell based
https://github.com/awesome-gocui/gocui

Todo:
- [x] Look at flush method used when editing item
- [x] CTRL+P for panel only works on second press
- [x] up/down doesn't work on content panel 
- [x] Check pageup/down functionality


lawrencegripper:
@stuartleeks Are you able to pull this branch and give it a go?

I've moved the `CTRL+P` for panel to `CTRL+Space` as there was an odd behaviour for me on Linux where it only registered on the second press (even with a normal keyboard).

I'm keen to move over to this one now I've found out that it supports a [simulated screen for integration testing](https://github.com/gdamore/tcell#testability). Think we could use this to add more "real" tests combined with the existing dummy ARM server stuff used by the FileSystem tests.

One thing I've yet to tackle is handling updating. Would you be alright with it being that
1. Default to "built-in" azbrowse editing in the itemView panel
2. Configurable VSCode based editing 
This would mean we don't need to work with flushing etc... maybe this is me being lazy... call me out if it is!

lawrencegripper:
So think I've nearly got the existing editing terminal based experience working. If I modify the `tcell_driver.go` in the new fork to have methods which allow `Fini` and `Init` to be called on `tcell.Screen` then it looks like it works.

```
var screen tcell.Screen

func TcellInit() error {
	tcellInit()
}

func TCellClose() {
	screen.Fini()
}
```

Outstanding issue is that after returning to azb the next screen draw isn't triggered until a user presses a key so there is a blank terminal.

Working out if I can make this useful and avoid the blank screen.

stuartleeks:
Just giving it a spin...

- Not sure if this is changed behaviour or not, but when typing in the right-hand pane if you hit backspace you are navigate back up the tree. 
- tested editing an item with nano and it worked fine :-)
- I miss Ctrl+P!

Overall, it's looking good 👍

lawrencegripper:
It turns out the CTRL+P thing is only an issue on my box to do with the command prompt setup. I've changed the default back!

Last outstanding item before this goes in is to look at the backspace behaviour you saw @stuartleeks, I can repro and I agree it's not what we want. Going to move it from a `global` to a `listhandler` go it only happens on list nav.

PR #326, opened 2021-04-16
Enable dashboard editing

stuartleeks:
When a resource group is expanded, the `ResoureGroupExpander` is saved as the expander. Since this doesn't support `Update`, it blocked updating any top-level resources, i.e. those added by the `ResourceGroupExpander`.

When attempting an Update, this PR adds a test to see whether the `SwaggerExpander` can handle it if the item's expander can't. This could be generalised to check all handlers, but the the top-level resources in a resource group the `SwaggerExpander` feels like the most likely candidate .


lawrencegripper (approved):
LGTM

PR #333, opened 2021-04-21
Only allow single Refresh at a time

### LINKED ISSUE #315: Refreshing multiple times ends in incorrect place in tree
If I open a subscription and press `F5` then the display is refreshed. If I then press `F5` again then I end up back at the subscription list

stuartleeks:
Fixes #315 

- Prevent multiple concurrent refreshes (since refresh performs a Back and Navigate using the current list state)
- Disable UI updates during a refresh (to mask the Back/Navigate actions)

### DIFF
--- a/internal/pkg/views/list.go
+++ b/internal/pkg/views/list.go
@@ -32,6 +32,8 @@ type ListWidget struct {
 	// To avoid blocking the "UI" thread opening items is done in a go routine. This lock prevents duplicates.
 	navLock      sync.Mutex
 	isNavigating bool
+	refreshLock  sync.Mutex
+	isRefreshing bool
 }
 
 // ListNavigatedEventState captures the state when raising a `list.navigated` event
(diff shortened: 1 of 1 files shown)


internal/pkg/views/list.go
```diff
@@ -219,19 +233,32 @@ func (w *ListWidget) Refresh() {
 	}
 
 	w.GoBack()
-	w.ExpandCurrentSelection()
+	w.expandItem(currentExpandedItem)
 
-	w.ChangeSelection(currentSelection)
+	// wait for navigation before resetting previous selection
+	go func() {
+		navigatedChannel := eventing.SubscribeToTopic("list.navigated")
```
lawrencegripper, line 240:
```suggestion
		defer errorhandling.RecoveryWithCleanup()
		navigatedChannel := eventing.SubscribeToTopic("list.navigated")
```

In the off chance that this fails adding recovery will ensure the terminal is left in a usable state and stack is printed etc

lawrencegripper: Think this'll need an additional import of `"github.com/lawrencegripper/azbrowse/internal/pkg/errorhandling"`

lawrencegripper (approved):
LGTM

PR #347, opened 2021-05-12
Move mouse support to '--mouse' option

lawrencegripper:
- Mouse support behind `--mouse`. Default to disabled
- Move to latest gocui


lawrencegripper:
This resolves #347, until we're able to support both mouse clicks and text selection, mouse support will be off by default and enabled with `--mouse`

PR #355, opened 2021-05-25
Support custom cross-subscription views with Azure Resource Graph

### LINKED ISSUE #306: cross-subscription search
I often know the name of a resource, but I couldn't tell which subscription it's in, (one of the many test subs picked by our CI at random)

in the portal, I can paste the resource name, and it will search in all subscription (or pre-selected subscriptions)

I don't think there is a way to do this in azbrowse, and it would be a nice addition.

lawrencegripper:
The basics of this are you write a `~/.azbrowse-query.config` file which is in the following format:

```
# 5 items
resourcecontainers 
| where type == 'microsoft.resources/subscriptions/resourcegroups' 
| take 5
---
# anotherquery
resourcecontainers | where type == 'microsoft.resources/subscriptions/resourcegroups' | take 5
```

(I went for a custom format for the query file as jamming these queries in JSON or YAML felt nasty - not sure if this was a good idea)

These are then loaded at the end of the `subscriptions` list. When expanded they show a list of all the resource groups returned by the query. 

image

The query MUST return a set of resource groups, hence the `where` clause in the example above.

Closes #306 

@serbrech I think this meets your brief, you can write a query that returns all resource groups you have access too then use the `/` filter to search through them for the one you want.

I'm still playing with the format of the file and how the UX flows together, feedback very much welcome. If you want to try it out checkout the branch, create the example file then run `make run` and scroll to the bottom, you queries should be there.


lawrencegripper:
Tagged a few reviewers as I'd like to get input on how this is tied together.

stuartleeks:
Ah, interesting. When I first scanned this I thought this was a more generic ability to run Azure Resource Graph queries!

Sounds cool - I'll try to find some time to give it a spin.

As a future step, it might be interesting to see whether there's a way to match the `type` against the swagger spec data to handle arbitrary resource types. Being able to define and run a query that way would open up some interesting options (and might be an alternative approach to the 'favourites' idea that has been discussed previously)

martinpeck:
Would "1 file per query" be better, and have each file be a .kql file containing a single kusto query?

- Maybe define a folder (`~/.azbrowse/`?) where they all live, or read them from the working directory/startup params
- 1 file per query makes it easier to share those queries, and also to drop them into a project's git repo to share with other devs
- "name of file" can be "name of query", so that `my-query.kql` results in `my query` as the name shown in azbrowse
- use `.kql` as the file extension as that appears to be the file extension for Kusto generally used
- your code then just assumes everything inside the file is a kusto query, rather than having to deal with special formatting. i.e. no need for any of the special formatting, the `---` can go, the query names can go


e.g.

### anotherquery.kql
``` text
// comments in kusto queries are fine
resourcecontainers 
| where type == 'microsoft.resources/subscriptions/resourcegroups' 
| take 5
```

lawrencegripper:
@martinpeck this is a much nicer way of handling this, going to go and give it a go now. Thankyou! 🙌

@stuartleeks interesting, I might be able to make this work, love the idea you could create a view which is a grouping of 'all AKS clusters from any RG' etc

stuartleeks:
@lawrencegripper - yeah, that sort of query feels like it would be handy. Or querying all resources with a specific tag. Feels like it would open up a number of cool options :-)

serbrech:
So... I could have a query that returns all resources under all subscriptions in a flat list, and then the filter would behave the way I expect it to?

Why isn't that the default instead of forcing a hierarchical navigation?

lawrencegripper:
> So... I could have a query that returns all resources under all subscriptions in a flat list, and then the filter would behave the way I expect it to?

Yeah, hopefully when I've finished this work, in its current state you can flatten all RG's over multiple subs

> Why isn't that the default instead of forcing a hierarchical navigation?

Azure Resource Graph didn't exist when I first built AzBrowse, then it did exist but the lag time for ingest was too high, then it didn't support Resource Group queries (`resourceContainers`) and then it's now where I took another look and it supported them and seemed to have less lag.

(Not sure that it was meant to be but came across to me as a little hostile there.)

lawrencegripper:
Need to work out how to handle duplicates in the list. 
```
Sub: A -> RG: Bob
Sub B -> RG: Bob
```
Currently, these two are indistinguishable from each other. Given subs group resources for `prod`/`staging` it feels like not being able to easily tell which sub returned the result is quite dangerous. 

Am I over thinking this, guess it can be up to the user to write their query appropriately as to prevent this being an issue?

stuartleeks:
If we're allowing cross sub querying then does it make sense to include the sub name in the node text somehow?

lawrencegripper:
Yeah that would be nice, sadly is looks like all that is included in `resourcecontainers` table is the subscriptions guid. 😔

Guess the code could make an additional query to build up a `SubId -> Name` map then use that to populate it.

Tried this out the following query `resourcecontainers | where type == 'microsoft.resources/subscriptions'` combined with the user query would let me call out the subs name of each resource. I like it - thanks @stuartleeks awesome idea!

stuartleeks:
Another idea for future is to add a cmdline switch that specifies a query folder to load and then use that to populate the root nodes on startup. Could be cool for defined deployments/projects/environments? (And would also give @serbrech a way to replace his default experience 😉)

stuartleeks:
Another quick thought - it might be cool if the "sort items" action sorted on subscription then resource?

lawrencegripper:
> Another idea for future is to add a cmdline switch that specifies a query folder to load and then use that to populate the root nodes on startup.

This is a cool idea so `--query-folder ./prod-1-queries` would do what @serbrech wants. I like it.

> it might be cool if the "sort items" action sorted on subscription then resource

So I've been thinking on this one a bit and as Kusto supports `| orderby xxx` I'm tempted to say this is up to the query author and azb renders them as they are returned. That way if folks want a certain ordering they can do it in the query based on whatever features they want (tags, sub, combinations etc)

serbrech:
No hostility here, just quick feedback from the phone with not enough care for form. Sorry.
This looks absolutely awesome :), and if it performs well, should just become the default imo.

Is that what portal uses when performing searches across subs?

Regarding sub names/ids.

Sub names can be long. Sub ids have constant length.
I wouldn't mind seeing the sub ID instead of the name personally

lawrencegripper:
Right think v1 of this is good to go. 

You can now write queries that return a bunch of resources or resource groups. The `subscriptionId` and `subscriptionName` are shown in the UI alongside the `resourceType`. There is even a shiny new doc on how to use this. 

There is no `--query-folder ./prod-1-queries` param as this is already handled by the `--navigate` param. If you have a query file call `sqlServers.kql` then `azbrowse --navigate sqlServers.kql` will open you straight to these query results.

lawrencegripper:
Hopefully merge this shortly then would be good to track feedback on https://github.com/lawrencegripper/azbrowse/issues/306 once people have had a play.

PR #358, opened 2021-06-15
Adding MS Graph ops

damoodamoo:
This adds some MS Graph operations - 
- Me: Shows the current user profile
- Apps: Shows list of apps
    - Search by name
    - Get by ID
        - Update App
        - View Owners
    - New App
    - Owned Apps
    - Deleted Apps
- Service Principals
    - Search by name
    - Get by ID

The `MS Graph` menu option is manually added to the list of subs - doesn't feel ideal but since the Graph is separate from a sub it felt like the only place. 

A few tweaks were made to the `armClient` to support the graph endpoint, getting a token from the CLI using `--resource-type ms-graph` (who knew?!). This will probably want finessing as it's a bit hacky.

CC @stuartleeks


lawrencegripper:
Oh cool - thanks!!

 I’ll take a look once I’m back from holiday

lawrencegripper:
The build error looks like ours my bad rather than yours, I’ll take a look at it separately. Looks like ours not correctly handling that this is a PR and will be missing secrets

lawrencegripper:
I've got a PR in now which resolves the issue you see with the build 🤞

https://github.com/lawrencegripper/azbrowse/pull/359

Once it's green I'll merge into main so you can pull into this PR.

internal/pkg/expanders/graph.go
```diff
@@ -0,0 +1,595 @@
...
+		SourceDescription: "GraphExpander failed to show app or SP",
+		Err:               err,
+	}
+}
+
+func (e *GraphExpander) newApp(ctx context.Context, currentItem *TreeNode) ExpanderResult {
+
+	// get the data for the editor panel
+	data := "{ \"displayName\": \"Replace this\" }"
+	content, err := editor.OpenForContent(data, ".json")
```
lawrencegripper, line 485: Nice!

internal/pkg/expanders/graph.go
```diff
@@ -0,0 +1,595 @@
...
+		return e.showMe(ctx, currentItem)
+	case itemTypeSpList:
+		return e.listAppsOrSps(ctx, currentItem, "")
+	case itemTypeAppList:
+		return e.listAppsOrSps(ctx, currentItem, "")
+	case itemTypeAppOrSp:
+		return e.showAppOrSp(ctx, currentItem)
+	case "graph":
+		return e.listRootMenu(ctx, currentItem)
+	case "action":
```
lawrencegripper, line 105: Is this required? (Honestly a question, slightly rusty on how actions hook up)

My recollection is that `ListActions` and `ExecuteActions` are going to handle this and `Expand` only matters for treenodes.

internal/pkg/expanders/graph.go
```diff
@@ -0,0 +1,595 @@
...
+
+	return ExpanderResult{
+		SourceDescription: "GraphExpander failed to list apps owned by ME",
+		Err:               err,
+	}
+}
+
+func (e *GraphExpander) getAppByID(ctx context.Context, currentItem *TreeNode) ExpanderResult {
+	commandChannel := make(chan string, 1)
+	commandPanelNotification := func(state interfaces.CommandPanelNotification) {
```
lawrencegripper, line 372: Optional: This method seems pretty useful maybe a good candidate to move under the `commandPanel` for reused. 

So `appID := <-commandChannel` would become `appID := <- e.commandPanel.CompletedInput()` or something

lawrencegripper (approved):
This looks awesome and has lots of very cool features, great work @damoodamoo

Once you've finished up the code could you record a demo of the features and add it to `docs/getting-started.md` under `Working with Microsoft Graph` heading? No worries if you don't get round to that bit. It's a nice to have.  

Chore: two old tests need a new `nil` parameter here: 

```
  internal/pkg/filesystem/filesystem_test.go:104:53: too few arguments in call to expanders.InitializeExpanders (typecheck)
  	expanders.InitializeExpanders(client, nil, nil, nil)
  	                                                   ^
  internal/pkg/views/notifications_test.go:71:53: too few arguments in call to expanders.InitializeExpanders (typecheck)
  	expanders.InitializeExpanders(client, nil, nil, nil)
  	                                                   ^
```

internal/pkg/expanders/graph.go
damoodamoo: I have it there because I was previously seeing the root menu get re-displayed underneath search results :/

damoodamoo:
Ok tests have been fixed up so hopefully should be good to go now!

lawrencegripper:
🚀 Nice!

PR #391, opened 2021-11-24
Add config to handle static web app POST-as-GET endpoints

stuartleeks:
- Add config to handle static web app POST-as-GET endpoints
- make swagger-codegen


lawrencegripper (approved):
LGTM, Thanks for fixing up!

PR #393, opened 2021-12-01
Compare Filter using ToLower

martinpeck:
This makes filtering case insensitive. 

Currently, the items in the list have ToLower applied, but the filter itself doesn't. This results in a filter of `M` not finding anything, my `m` will find items that include both upper and lower case `m` characters.

### DIFF
--- a/internal/pkg/views/list.go
+++ b/internal/pkg/views/list.go
@@ -78,7 +78,7 @@ func (w *ListWidget) SetFilter(filterString string) {
 	}
 	filteredItems := []*expanders.TreeNode{}
 	for _, item := range w.currentPage.Items {
-		if strings.Contains(strings.ToLower(item.Display), filterString) {
+		if strings.Contains(strings.ToLower(item.Display), strings.ToLower(filterString)) {
 			filteredItems = append(filteredItems, item)
 		}
 	}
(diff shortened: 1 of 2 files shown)


martinpeck:
I considered using https://pkg.go.dev/strings#EqualFold, but this change is in keeping with the rest of the codebase.

internal/pkg/views/list.go
```diff
@@ -78,7 +78,7 @@ func (w *ListWidget) SetFilter(filterString string) {
 	}
 	filteredItems := []*expanders.TreeNode{}
 	for _, item := range w.currentPage.Items {
-		if strings.Contains(strings.ToLower(item.Display), filterString) {
+		if strings.Contains(strings.ToLower(item.Display), strings.ToLower(filterString)) {
```
stuartleeks, line 81: Could you create `filterStringLower` outside the `for` loop to avoid performing the conversion multiple times?

stuartleeks (changes requested):

internal/pkg/views/list.go
martinpeck: Good call. I've changed this instance, and the other instance where a filter is used in a loop

stuartleeks (approved):
LGTM - thanks!

PR #408, opened 2022-07-04
Handle missing codeblock finish in readme

lawrencegripper:
Marketplace catalog readme has a YAML block with missing end closing block. This makes swagger update handle this by skipping the badly formed version.

### DIFF
--- a/scripts/swagger_update/api_set.py
+++ b/scripts/swagger_update/api_set.py
@@ -118,10 +118,10 @@ def get_api_version_tag(resource_provider_name, readme_contents, overrides):
 
 code_block_end_regex = re.compile("^[\\s]*```[\\s]*$", flags=re.MULTILINE)
 def find_api_version(resource_provider_name, readme_contents, version_tag, input_file_additions):
-
     # Regex to match:   ```yaml $(tag) == 'the-version-tag`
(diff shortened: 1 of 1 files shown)


scripts/swagger_update/api_set.py
```diff
@@ -118,10 +118,10 @@ def get_api_version_tag(resource_provider_name, readme_contents, overrides):
 
 code_block_end_regex = re.compile("^[\\s]*```[\\s]*$", flags=re.MULTILINE)
 def find_api_version(resource_provider_name, readme_contents, version_tag, input_file_additions):
-
     # Regex to match:   ```yaml $(tag) == 'the-version-tag`
     # Also match:       ```yaml $(tag) == 'the-version-tag` || $(tag) == 'some-other-tag'
     # But don't match   ```yaml $(tag) == 'the-version-tag' && $(another-condition)
+    print(version_tag)
```
lawrencegripper, line 124:
```suggestion
```

scripts/swagger_update/api_set.py
```diff
@@ -160,6 +163,7 @@ def get_api_version_from_readme(resource_provider_name, readme_path, version_ove
         print("==> no version tag found in readme: " + readme_path)
         return None
 
+    print(readme_path)
```
lawrencegripper, line 166:
```suggestion
```

PR #431, opened 2023-02-11
feat: Remove direct calls to az cli to create access tokens

tmeckel:
This PR contains changes which replaces all direct calls to AzCLI to create access tokens, with calls to `github.com/Azure/azure-sdk-for-go/sdk/azidentity.NewDefaultAzureCredential` .

This will provide support for authenticating via environment variables, managed identities and calling AzCLI as last resort, if none of the aforementioned possibilities were successful.

image

https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash

**Authenticating via:**
1. [Service principal with a secret](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#service-principal-with-a-secret)
2. [Service principal with certificate](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#service-principal-with-certificate)
3. [Username and password](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#username-and-password)
4. [Use a managed identity](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#-option-2-use-a-managed-identity)
5. [Sign in with Azure CLI](https://learn.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash#-option-3-sign-in-with-azure-cli)


lawrencegripper:
Hi, Thanks for the PR :+1: 

History wise I got steered away from an original solution to reach manually into `.azure` json file to grab tokens because this wouldn't handle refreshes and the format wasn't something I could depend on.

It looks like this library takes away that pain, assuming we keep it up-to-date, sound about right? [Looks like Azure Terraform Provider has made a similar shift.](https://github.com/hashicorp/terraform-provider-azurerm/blob/2e3faf9990d7740e02159ce18feb864477d9c1f4/website/docs/guides/azure_cli.html.markdown?plain=1#L9)

Going to give this a test now.

tmeckel:
@lawrencegripper I think I fixed the linting issues. Thanks for pointing me to `make checks`. Overlooked it in the first place.

lawrencegripper:
Thanks :heart: 

I've fallen behind on some tooling updates which are tripping up the CI for this one, hope you don't mind if I push some updates to get this :green_circle: 

Going to tackle the non-essential bits in 
- https://github.com/lawrencegripper/azbrowse/pull/432

lawrencegripper:
Past Lawrence didn't make this easy, don't like that guy at the moment.

lawrencegripper (approved):
CI's now behaving. LGTM let's ship ⛴️ 

Thanks for all the hard work on the PR 🙇‍♂️

PR #433, opened 2023-02-17
Move to using devcontainer ci for build

lawrencegripper:
Move to using devcontainer CI action to run builds using the devcontainer definition
- https://github.com/devcontainers/ci

Use devcontainer/cli to allow folks to run without VSCode 
- https://github.com/devcontainers/cli/


.github/workflows/build.yaml
```diff
@@ -1,4 +1,7 @@
 name: release
+permissions:
+  packages: write
+  contents: read
```
lawrencegripper, line 4: Remove these as passing the token in from secrets

.github/workflows/build.yaml
```diff
@@ -8,30 +11,47 @@ on:
...
 
-    - name: Set up Docker BuildKit
-      uses: docker/setup-buildx-action@v1
-
-    - name: Build devcontainer for tooling
-      run: make devcontainer
-
-    - name: Build the Docker image, run tests and publish(on master only)
-      run: CMD='ruby ${PWD}/scripts/release.rb' make devcontainer-run
+    - name: Pre-build image and run make ci-build in dev container
```
lawrencegripper, line 28: Name this better

.github/workflows/build.yaml
lawrencegripper:
```suggestion
```

.github/workflows/build.yaml
lawrencegripper:
```suggestion
    - name: Run Checks, Tests and Build
```

PR #579, opened 2024-07-25
Fix swagger update

stuartleeks:
- Force `ai/dataplane` version to avoid picking up version that doesn't have spec files
- Copy `common-types` folders to fix mysql codegen error
- make swagger-update
- make swagger-codegen

NOTE: the number changes is quite large (as is typical with swagger updates). I've split out commits for code changes separately from the swagger update/codegen to make it easier to find and review those changes


lawrencegripper (approved):
:+1: LGTM

PR #580, opened 2024-07-29
Add support for viewing container app logs

stuartleeks:
Add support for viewing container app logs
- support drilling into the containers under a revision's replica
- add command to view logs for a container
- support streaming logs and cancel context to terminate log request when user navigates away from the log view

Also updates the dev container to include `GOPATH/bin` in the `PATH` so that `azbrowse` is found in the release script after running `make install`

### DIFF
--- a/internal/pkg/expanders/containerApp.go
+++ b/internal/pkg/expanders/containerApp.go
@@ -235,26 +235,6 @@ func (e *ContainerAppExpander) GetAuthToken(ctx context.Context, currentItem *Tr
 	return authTokenResponse.Properties.Token, nil
 }
 
-// func (e *ContainerAppExpander) getLogs(ctx context.Context, logStreamEndpoint string, authToken string) (string, error) {
-// 	// TODO - actually get the logs!
-
-// 	request, err := http.NewRequestWithContext(ctx, "GET", logStreamEndpoint, nil)
-// 	if err != nil {
-// 		return "", fmt.Errorf("failed to create request: %s", err)
-// 	}
-// 	request.Header.Set("Authorization", "Bearer "+authToken)
-
-// 	httpClient := http.DefaultClient
-
-// 	response, err := httpClient.Do(request)
-// 	if err != nil {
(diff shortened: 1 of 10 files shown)


internal/pkg/expanders/containerApp.go
```diff
@@ -0,0 +1,286 @@
...
+
+	var authTokenResponse ContainerAppAuthTokenResponse
+	err = json.Unmarshal([]byte(data), &authTokenResponse)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse auth token response: %s", err)
+	}
+	return authTokenResponse.Properties.Token, nil
+}
+
+// func (e *ContainerAppExpander) getLogs(ctx context.Context, logStreamEndpoint string, authToken string) (string, error) {
```
lawrencegripper, line 238: Think this can be removed now right? As the command handler is doing it

lawrencegripper (commented):
Looks really nice. (I need to get the code out again and remember how everything is wired up)

Other than commented out method lgtm

internal/pkg/keybindings/list_containerapp_logs.go
```diff
@@ -0,0 +1,134 @@
...
+		// Cancel log context
+		cancel()
+	}()
+
+	return h.getLogs(ctx, logStreamEndpoint, authToken)
+
+}
+
+func (h *CommandPanelContainerAppLogsHandler) getLogs(ctx context.Context, logStreamEndpoint string, authToken string) error {
+	// TODO - actually get the logs!
```
lawrencegripper, line 95: Think this is now done

internal/pkg/expanders/containerApp.go
stuartleeks: Oops - I meant to remove this!

internal/pkg/keybindings/list_containerapp_logs.go
stuartleeks: lol - good spot, thanks!

lawrencegripper (approved):
Lgtm

PR #586, opened 2024-08-06
Add fuzzy filtering

stuartleeks:
- [X] Add a new fuzzy filter command for fuzzy string matching (e.g.  `srnm` would match `some resource name`)
- [X] Allow setting default key bindings with modifiers. Using `/` for fuzzy filter (compared to `Alt+/` for non-fuzzy filtering)
- [X] Allow uesr-configured key bindings to include modifiers


internal/pkg/keybindings/defaults.go
```diff
@@ -39,6 +39,7 @@ var DefaultKeys = map[string]interface{}{
 	"commandpanelup":      gocui.KeyArrowUp,
 	"commandpanelenter":   gocui.KeyEnter,
 	"filter":              rune('/'),
```
lawrencegripper, line 41: I wonder if we should take over this one and make fuzzy filter the default

internal/pkg/keybindings/global_filter_commandpanel.go
```diff
@@ -11,16 +11,22 @@ type CommandPanelFilterHandler struct {
 	commandPanelWidget *views.CommandPanelWidget
```
lawrencegripper, line 11: Classic me typo on file name, thanks 🙏

lawrencegripper (approved):
Love it 🥰

internal/pkg/views/list.go
```diff
@@ -113,15 +121,44 @@ func (w *ListWidget) itemsToShow() []*expanders.TreeNode {
...
+					// we're done, add any remaining characters
+					if displayIndex+1 < len(displayText) {
+						result += displayText[displayIndex+1:]
+					}
+					break
+				}
+			} else {
+				// not a match - add this character (from the non-lowered display text)
+				result += string(displayText[displayIndex])
+			}
```
lawrencegripper, line 152: 👆 this is fun!

internal/pkg/keybindings/defaults.go
stuartleeks: Whichever is default, I quite like the idea of `Alt+/` (or `Shift+/`) as the other option 🤷

internal/pkg/views/list.go
stuartleeks: I'm open to suggestions for improvements. This got a little gnarly as I wanted case-insensitive matching but preserving the case in the output!

internal/pkg/keybindings/defaults.go
stuartleeks: Fuzzy filter now binds to `/` by default, and the non-fuzzy filter binds to `Alt+/`

internal/pkg/views/list.go
stuartleeks: I've made a change to the implementation of this that I think simplifies it slightly - let me know what you think :-)

lawrencegripper (approved):

PR #637, opened 2025-04-28
Upgrade Golang CI lint

lawrencegripper:
Seeing updated CI crash due to failure in golangci-lint so updating.


lawrencegripper:
Work needed there, the config was out of date to new format

